PR #8, opened 2017-01-20
Require double back press to close application

Tommy-Geenexus:
Prevent accidentially exiting the app by forcing a double back press in the specified time interval (default: 2s) to exit.


thestinger:
Is there a good reason to treat this app differently than the norm? Anything time-based doesn't seem great for accessibility and isn't the default fine? There aren't any local changes to lose.

Tommy-Geenexus:
If you accidentially hit the back button and leave the app, you will loose the currently loaded PDF, and need to reopen it. Thus i think it makes sense to prevent situations like these by forcing a double press.

Otherwise one could use a different method of saving app states so that the state would be preserved throughout app termination.

thestinger:
I think it makes more sense to save state if any approach is taken.

Tommy-Geenexus:
Alright, then i will close this. I'll try implementing that at some point in the future.

PR #9, opened 2017-02-10
Enable/disable menu items based on zoom-level

Tommy-Geenexus:
This patch disables and greys out listed menu items when min/max zoom level is reached, to show the user no further zooming is possible.

### DIFF
--- a/app/src/main/java/co/copperhead/pdfviewer/PdfViewer.java
+++ b/app/src/main/java/co/copperhead/pdfviewer/PdfViewer.java
@@ -21,7 +21,10 @@ import android.widget.FrameLayout;
 import android.widget.NumberPicker;
 
 public class PdfViewer extends Activity {
+    private static final int MIN_ZOOM_LEVEL = 0;
     private static final int MAX_ZOOM_LEVEL = 4;
+    private static final int ALPHA_LOW = 130;
+    private static final int ALPHA_HIGH = 255;
     private static final int ACTION_OPEN_DOCUMENT_REQUEST_CODE = 1;
     private static final String STATE_URI = "uri";
     private static final String STATE_PAGE = "page";
@@ -33,6 +36,8 @@ public class PdfViewer extends Activity {
     private int mNumPages;
     private int mZoomLevel;
     private Channel mChannel;
+    private boolean mZoomInState = true;
+    private boolean mZoomOutState = true;
 
     private class Channel {
         @JavascriptInterface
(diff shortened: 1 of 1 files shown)


app/src/main/java/co/copperhead/pdfviewer/PdfViewer.java
```diff
@@ -21,18 +21,24 @@
...
     private static final String STATE_ZOOM_LEVEL = "zoomLevel";
 
+    private boolean mButtonDisabled;
     private WebView mWebView;
     private Uri mUri;
     private int mPage;
     private int mNumPages;
     private int mZoomLevel;
     private Channel mChannel;
+    private MenuItem mIn;
```
thestinger: It would be better to call these mZoomIn and mZoomOut to be clearer.

app/src/main/java/co/copperhead/pdfviewer/PdfViewer.java
```diff
@@ -21,18 +21,24 @@
...
+    private static final int MIN_ZOOM_LEVEL = 1;
     private static final int MAX_ZOOM_LEVEL = 4;
+    private static final int ALPHA_LOW = 130;
+    private static final int ALPHA_HIGH = 255;
     private static final int ACTION_OPEN_DOCUMENT_REQUEST_CODE = 1;
     private static final String STATE_URI = "uri";
     private static final String STATE_PAGE = "page";
     private static final String STATE_ZOOM_LEVEL = "zoomLevel";
 
+    private boolean mButtonDisabled;
```
thestinger: mZoomButtonDisabled would be clearer, but I think you can do without this field. It has the information it needs without tracking the state.

app/src/main/java/co/copperhead/pdfviewer/PdfViewer.java
```diff
@@ -122,6 +128,30 @@ private void openDocument() {
...
+        if (!enable) {
+            item.setEnabled(false);
+            item.getIcon().setAlpha(ALPHA_LOW);
+        } else {
+            item.setEnabled(true);
+            item.getIcon().setAlpha(ALPHA_HIGH);
+        }
+    }
+
+    private void checkButtonEnableDisable(MenuItem item, int zoomlevel, int desired_level) {
```
thestinger: This is going to need to be done when restoring the zoom level from savedInstanceState too. You can test by changing the orientation of the app by rotating the device, which will cause the activity to restart from saved state. This should probably just check the current zoom level and handle enabling / disabling both buttons. There will end up being other forms of zooming that aren't single steps up and down.

Tommy-Geenexus:
@thestinger 

Pr has been updated:
- Code got cleaned up (the menu item handling has been moved to onPrepareOptionsMenu() as per    android menu guide suggestion)
- menu item handling works properly with instance state changes

One limitation though:
This only works on the menu items, not if they are displayed in the action bar as buttons, because onPrepareOptionsMenu() only gets invoked by menu items shown as list.
Its possible to make this work on buttons too, one could call invalidateOptionsMenu() to force update the options menu. Tell me wheter or not this is needed.

thestinger:
@Tommy-Geenexus It would be best to make it work on the buttons too, since they can be displayed if the screen is wide enough. I think it will display up to five right now, but it would make sense for it to show more in the future.

Tommy-Geenexus:
@thestinger 

Pr has been updated:
- item enabling/disabling now properly works on buttons and list items

I had to expand the code and implement methods to save/restore the button state,
because upon calling invalidateOptionsMenu() the menu got recreated and button states resetted to default state (enabled).

thestinger:
Thanks, applied as https://github.com/CopperheadOS/platform_packages_apps_PdfViewer/commit/c75788dd5b49eebffe549bf3ba3b089f0ed665d6 (GitHub web interface conflict resolution is dumb).

PR #10, opened 2017-02-17
Do not process menu item selections when no PDF is loaded

Tommy-Geenexus:
Theres no need to process any item selections except for opening a document on initial app launch.
(Also deals with: "Uncaught ReferenceError: onRenderPage is not defined")

### DIFF
--- a/app/src/main/java/co/copperhead/pdfviewer/PdfViewer.java
+++ b/app/src/main/java/co/copperhead/pdfviewer/PdfViewer.java
@@ -30,6 +30,10 @@ public class PdfViewer extends Activity {
     private static final int ALPHA_LOW = 130;
     private static final int ALPHA_HIGH = 255;
     private static final int ACTION_OPEN_DOCUMENT_REQUEST_CODE = 1;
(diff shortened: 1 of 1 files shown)


thestinger:
`mUri` is set before the page is actually loaded, so this will need to be toggled via a callback: https://developer.android.com/reference/android/webkit/WebViewClient.html#onPageFinished(android.webkit.WebView, java.lang.String). I think it's probably before to disable and dim all of the non-functional actions at the start rather than producing errors, and then enable them when the page is ready. It will need to disable them again when triggering another load too.

Tommy-Geenexus:
Thats what i thought of first, thought i dropped it because it would be harder to maintain.
But its the proper non half-baked way, so i'll go ahead and implement this hopefully soon.

Tommy-Geenexus:
@thestinger 
Pr updated:
- All Items except "open document" are dimmed at first
- After page has finished loading, all will get enabled
- When reload with new document is triggered, items will be disabled and enabled after page has been loaded

thestinger:
Thanks!

PR #11, opened 2017-02-21
Handle number picker in fragment

Tommy-Geenexus:
Let dialog lifecycle be handled by the dialog fragment.
This prevents window leakage upon device rotation  (which is currently happening: "... Activity co.copperhead.pdfviewer.PdfViewer has leaked window DecorView ...") and also saves/restores the dialog's instance state (which is currently neither saved nor restored).

PR #12, opened 2017-02-22
Show actual/max page numbers on page change

### LINKED ISSUE #2: page numbers

Tommy-Geenexus:
Closes #2 
Current page number/max page number are displayed at the bottom right
via a rectangle dark grey on white toast with some padding
(This is almost the same approach Google chose for their PDF App).

Note: This needs to be rebased after #11 gets merged to prevent a conflict.


thestinger:
Thanks, merged. It will be #11 that needs to be rebased because I need to think about it.

PR #14, opened 2017-05-28
Czech translation

mople71:
Hello! Won't hurt, right?


thestinger:
Hi, thanks for this.

I'm not quite sure how we want to handle translations yet so I'll need to sit on this for now. It's an issue because it's difficult for me to review translations for languages I don't know (i.e. not French) and if changes are made I can't update it myself.

thestinger:
I'll revisit this at a later date when I've figured out how we want to handle translations.

PR #16, opened 2017-07-01
Add option to show document properties

Tommy-Geenexus:
If you have any suggestions on code improvements, ping me.

### DIFF
--- a/app/src/main/java/co/copperhead/pdfviewer/PdfViewer.java
+++ b/app/src/main/java/co/copperhead/pdfviewer/PdfViewer.java
@@ -4,9 +4,14 @@ import android.app.Activity;
 import android.annotation.SuppressLint;
 import android.content.Intent;
 import android.content.res.ColorStateList;
+import android.database.Cursor;
 import android.graphics.Color;
+import android.graphics.drawable.Drawable;
 import android.net.Uri;
+import android.os.AsyncTask;
 import android.os.Bundle;
+import android.provider.OpenableColumns;
+import android.util.Log;
 import android.view.Gravity;
 import android.view.Menu;
 import android.view.MenuInflater;
@@ -21,10 +26,25 @@ import android.webkit.WebViewClient;
 import android.widget.TextView;
 import android.widget.Toast;
 
+import org.json.JSONException;
+import org.json.JSONObject;
+
 import java.io.InputStream;
 import java.io.IOException;
+import java.math.RoundingMode;
+import java.text.DecimalFormat;
+import java.text.SimpleDateFormat;
+import java.util.Calendar;
+import java.util.Date;
+import java.util.Locale;
+import java.util.TimeZone;
+
+import co.copperhead.pdfviewer.fragment.DocumentPropertiesFragment;
+import co.copperhead.pdfviewer.fragment.JumpToPageFragment;
 
 public class PdfViewer extends Activity {
+    private static final String TAG = "PdfViewer";
+
     private static final int MIN_ZOOM_LEVEL = 0;
     private static final int MAX_ZOOM_LEVEL = 4;
     private static final int ALPHA_LOW = 130;
@@ -40,13 +60,14 @@ public class PdfViewer extends Activity {
 
     private WebView mWebView;
     private Uri mUri;
-    int mPage;
-    int mNumPages;
+    public int mPage;
+    public int mNumPages;
     private int mZoomLevel = 2;
     private int mDocumentState;
     private Channel mChannel;
     private boolean mZoomInState = true;
     private boolean mZoomOutState = true;
+    private String mDocumentProperties;
     private InputStream mInputStream;
     private TextView mTextView;
     private Toast mToast;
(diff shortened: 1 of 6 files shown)


Tommy-Geenexus:
@thestinger 

I've updated the parseDate() method.
Malformed date metadata would lead to crashes,
so i've played it safe to avoid anything alike.
May not be the most elegant, but it works (now behaves just like the PDF.js demo).
Still doesn't sanity check dates like few other readers, but not sure if that's really necessary.
Please tell me if you're ok with this.

app/src/main/java/co/copperhead/pdfviewer/PdfViewer.java
```diff
@@ -40,13 +61,14 @@
 
     private WebView mWebView;
     private Uri mUri;
-    int mPage;
-    int mNumPages;
+    public int mPage;
```
thestinger, line 63: No need to mark these as public, the default without a visibility marker is package-private which is a bit nicer than public although here it doesn't really make any difference.

app/src/main/java/co/copperhead/pdfviewer/PdfViewer.java
```diff
@@ -211,7 +358,7 @@ private void enableDisableItems(Menu menu, boolean disable) {
         }
     }
 
-    void positiveButtonRenderPage(int selected_page) {
+    public void positiveButtonRenderPage(int selected_page) {
```
thestinger: Same here, no need to mark this public.

app/src/main/java/co/copperhead/pdfviewer/PdfViewer.java
thestinger: Ah, I see, you did the other thing in a separate package.

app/src/main/res/values/strings.xml
```diff
@@ -8,4 +8,20 @@
     <string name="action_zoom_out">Zoom out</string>
     <string name="action_zoom_in">Zoom in</string>
     <string name="action_jump_to_page">Jump to page</string>
+    <string name="action_view_document_properties">Document Properties</string>
```
thestinger: For consistency, these should be "Document properties", etc. rather than title case.

app/src/main/java/co/copperhead/pdfviewer/PdfViewer.java
Tommy-Geenexus: I've moved all fragments to a separate package.
In the future once this project expands it's better to keep things sorted, for better overview.
In case more activities will be added, i'd also suggest creating a separate package for them.

app/src/main/res/values/strings.xml
Tommy-Geenexus: Will update that in a minute.

Tommy-Geenexus:
Pr updated, adjusted spelling.

app/src/main/java/co/copperhead/pdfviewer/PdfViewer.java
```diff
@@ -66,6 +88,125 @@ public int getZoomLevel() {
...
+
+                        final String propertyFileName = getString(R.string.document_properties_file_name);
+                        final String fileName = cursor.getString(fileNameIndex);
+                        final String propertyFileSize = getString(R.string.document_properties_file_size);
+                        final String fileSize = parseFileSize(Long.valueOf(cursor.getString(fileSizeIndex)));
+
+                        cursor.close();
+
+                        final JSONObject json = new JSONObject(properties);
+                        final String propertyTitle = getString(R.string.document_properties_title);
```
thestinger: It seems better to change the first parameter of `formatProperty` to take an integer, call `getString` inside there and then pass these as parameters below without an extra variable for each one.

app/src/main/java/co/copperhead/pdfviewer/PdfViewer.java
thestinger: Yeah, that probably makes sense. I missed that when looking at this in isolation.

app/src/main/java/co/copperhead/pdfviewer/PdfViewer.java
```diff
@@ -66,6 +88,125 @@ public int getZoomLevel() {
...
+
+                private String parseFileSize(long fileSize) {
+                    final double kb = fileSize / 1024;
+                    if (((long) kb) == 0) {
+                        return String.format("%s Bytes", String.valueOf(fileSize));
+                    } else {
+                        final DecimalFormat format = new DecimalFormat("#.##");
+                        format.setRoundingMode(RoundingMode.CEILING);
+                        if (kb < 1024) {
+                            return String.format("%s KB (%s Bytes)", format.format(kb), String.valueOf(fileSize));
```
thestinger: Should be kB instead of KB (unlike MB).

app/src/main/java/co/copperhead/pdfviewer/PdfViewer.java
```diff
@@ -66,6 +88,125 @@ public int getZoomLevel() {
...
+
+        @JavascriptInterface
+        public void setDocumentProperties(final String properties) {
+            new AsyncTask<Void, Void, Void>() {
+                private String formatProperty(String property, String propertyValue) {
+                    return String.format("\n%s:\n%s\n", property, propertyValue);
+                }
+
+                private String parseFileSize(long fileSize) {
+                    final double kb = fileSize / 1024;
```
thestinger: I think this should use 1000, for consistency with the units elsewhere in Android's storage / file interfaces. Using 1024 isn't wrong but then the kB / MB below should be kiB / MiB per the SI definitions (despite the historical computer usage, which isn't standard).

Tommy-Geenexus:
Pr updated according to your suggestions.

Tommy-Geenexus:
@thestinger 

I've updated date parsing to match PDF v1.7 specs.
Now parses all possible dates and properly differentiates between valid and invalid ones,
plus shows dates according to specs when optional fields are missing.
(I've done this with JodaTime, default Java time libraries are a pita to work with)

thestinger:
Note that the build.gradle isn't used during the regular build process as part of AOSP. Ideally there won't be a new dependency.

Tommy-Geenexus:
I'm afraid that this dependency has to remain, as Java 8 support is only rudimentary and probably doesn't include new date libraries.
The issue is, with the Java 7 default libraries  that the provided date methods are not too well implemented and show weird behaviour at times, which got fixed with Java 8.

If it's really necessary i can try to rewrite, but can't the dependencies simply be moved to the top level build.gradle?

thestinger:
AOSP doesn't use gradle at all, it uses the Android.mk build system. If there's a new dependency it will need to be added as a repository to the CopperheadOS manifest and it needs to have an Android.mk / Android.bp build system. Android.mk is on the way out but I think for now it has to be Android.mk.

Tommy-Geenexus:
Ah yeah, right. Long time since i've built my last firmware.
So it depends, are you willing to add a repo to the manifest?
If not, the method must to be reworked.

thestinger:
If it's a pure Java library that might be okay, if it's trivial to build, but wouldn't want to do it if it has any native / JNI code. It might add a painful maintenance burden over time so it would really be ideal to avoid it. There might be a library in the Android Open Source Project tree that can be used?

thestinger:
The tricky part is keeping this as something that can be built in both AOSP (the main way for production) and outside of it via gradle. AOSP probably has something for this but it may not be available / used outside of AOSP.

Tommy-Geenexus:
I'll reimplement this with existing Java 7 Api then. Definitly not worth the extra hassle then.

thestinger:
Using https://developer.android.com/reference/android/text/format/package-summary.html seems fine if there's something useful in there.

Tommy-Geenexus:
Reimplementation is finished and extensively tested.
Library dependency has been removed at the cost of some uglier code, but that's the tradeoff.
The date parser should work correctly, tested with some modified sample PDF's.
I've also added some verbosity to log all parsing errors, should help with resolving potential issues / bugs.

Tommy-Geenexus:
@thestinger 
Fixed another issue with incorrect time zone in SimpleDateFormat.
Time correctly gets parsed as UTC now.
Should be all for this pr, any other changes to this (if at all) will be published in a follow up pr.

Additionally, i'll start working on the immersive mode in the coming weeks, just to let you know.

PR #18, opened 2017-07-06
Enable immersive mode for viewing PDF's

Tommy-Geenexus:
(Closes https://github.com/CopperheadOS/platform_packages_apps_PdfViewer/issues/5)
@thestinger 

<del>One (more or less) issue with the current implementation:
When a user triggers the clipboard prompt, then taps, the clipboard prompt will be hidden and the immersive mode will show/hide system ui all within the same tap.</del>

<del>On the google pdf viewer, you must tap once to close the clipboard prompt in case its open,
and a second time to show/hide system ui (which i think is appropriate).</del>

<del>Is there a good way to check whether or not the clipboard prompt is open/the user selected text?</del>

Nevermind, figured it out.


Tommy-Geenexus:
Patch is complete and ready for review

thestinger:
@Tommy-Geenexus Can you cherry-pick this to a new branch based on the new nougat-mr2-release branch and resubmit it as a pull request? I had to make a change to the history and I thought I'd be able to fix this for you but it won't allow that.

PR #20, opened 2017-08-18
Add support for encrypted PDF's

Tommy-Geenexus:


thestinger:
I switched this to track the oreo-r6-release branch. I won't do any major changes that could break it before I have time to deal with it but I did bump targetSdkVersion to 26 (not minSdkVersion at least until Nexus 9 is dropped and CopperheadOS stable is migrated to Oreo) and upgraded pdf.js.

thestinger:

app/src/main/java/co/copperhead/pdfviewer/PdfViewer.java
```diff
@@ -348,34 +394,68 @@ public void onPageFinished(WebView view, String url) {
...
+        loadPdf();
+        invalidateOptionsMenu();
+    }
+    
+    private void loadPdf() {
         mWebView.loadUrl("file:///android_asset/viewer.html");
     }
 
+    public void loadPdfWithPassword(String password) {
+        mWebView.evaluateJavascript(String.format("loadDocument(\"%s\")", password), null);
```
thestinger, line 447: This should really be changed to passing it via the channel. It's too hard to escape things properly to pass them as arguments via evaluateJavascript.

app/src/main/java/co/copperhead/pdfviewer/PdfViewer.java
```diff
@@ -41,27 +44,31 @@
 
 import co.copperhead.pdfviewer.fragment.DocumentPropertiesFragment;
 import co.copperhead.pdfviewer.fragment.JumpToPageFragment;
+import co.copperhead.pdfviewer.fragment.PasswordPromptFragment;
 
-public class PdfViewer extends Activity {
+public class PdfViewer extends AppCompatActivity {
```
thestinger, line 49: I'm okay with using AppCompat now since Google is essentially making the support library mandatory with Android P so that should make this easier. They're even going to be deprecating using native fragments, so there's no point in avoiding the support library.

thestinger: Their documentation is still presenting it as primarily for backwards compatibility but they're essentially hoisting everything they can out of the platform into it now.

app/src/main/java/co/copperhead/pdfviewer/PdfViewer.java
```diff
@@ -41,27 +44,31 @@
...
     private static final String TAG = "PdfViewer";
 
     private static final int MIN_ZOOM_LEVEL = 0;
     private static final int MAX_ZOOM_LEVEL = 4;
     private static final int ALPHA_LOW = 130;
     private static final int ALPHA_HIGH = 255;
-    private static final int ACTION_OPEN_DOCUMENT_REQUEST_CODE = 1;
+    private static final int ACTION_OPEN_DOCUMENT_REQUEST_CODE_1 = 1;
```
thestinger, line 56: These should probably have better names.

app/src/main/java/co/copperhead/pdfviewer/PdfViewer.java
Tommy-Geenexus: Escaping like that works fine based on my testing, i did the same thing with the option to rotate PDF's, and the only change needed was to parse the arg string to integer.
But if you insist, i'm going to change it.

Tommy-Geenexus:
@thestinger 
In general there are more things which need to be adressed, i.e the entered password is stored in shared prefs so that it musn't be entered again after rotation, but it's never flushed, and more.

I will take care of immersive mode first (there are some changes due), and then i'll start looking into the details on this one.

app/src/main/java/co/copperhead/pdfviewer/PdfViewer.java
thestinger: What if it contains a `"` character though?

Tommy-Geenexus:
Sorry my bad, i fucked up. Obviously it's stored in a bundle already (facepalm), so i can scrap that.

app/src/main/java/co/copperhead/pdfviewer/PdfViewer.java
Tommy-Geenexus: You're right about that (makes sense duh!), my test sample seemed to have ignored non alpha/digit's, i'll add that to my todo list.

PR #22, opened 2017-11-08
make Pdf Viewer behave like Android 5.0 "Document Task"

eighthave:
Android 5.0 started making document viewer apps behave like multiple windows.  This makes Pdf Viewer act like that.  It also makes Pdf Viewer listen to changes in the `ContentProvider` that it received files from other apps.  So if OpenKeychain locks, then Pdf Viewer will stop showing the PGP-encrypted document.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,6 @@
 .gradle
 local.properties
 build/
+
+.idea/
+*.iml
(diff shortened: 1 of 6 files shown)


eighthave:
added one fix to make sure the keyboard is hidden when launch, since OpenKeychain left it open when it prompts the user while decrypting.

thestinger:
I applied one of the small fixes but I haven't had time to look at the rest and unfortunately this conflicts with some other small fixes that I merged.

thestinger:
GitHub is doing something weird with the change log too.

thestinger:

PR #27, opened 2017-11-15
Refactor document properties and parsing

Tommy-Geenexus:
Since AsyncTask is an anonymous inner class, there is a possibility of leaking memory, which AS 3.0 warns about, thus it's been mitigated to an AsyncTaskLoader, which is a safe wrapper around AsyncTask, and handles configuration changes itself.

### DIFF
--- a/app/src/main/assets/viewer.js
+++ b/app/src/main/assets/viewer.js
@@ -188,6 +188,8 @@ PDFJS.getDocument("https://localhost/placeholder.pdf").then(function(newDoc) {
     channel.setNumPages(pdfDoc.numPages);
     pdfDoc.getMetadata().then(function(data) {
         channel.setDocumentProperties(JSON.stringify(data.info));
+    }).catch(function(err) {
+        console.log(err);
     });
     renderPage(channel.getPage(), false, false);
 });
--- a/app/src/main/java/co/copperhead/pdfviewer/PdfViewer.java
+++ b/app/src/main/java/co/copperhead/pdfviewer/PdfViewer.java
@@ -2,15 +2,13 @@ package co.copperhead.pdfviewer;
 
 import android.app.Activity;
 import android.annotation.SuppressLint;
+import android.app.LoaderManager;
 import android.content.Intent;
+import android.content.Loader;
 import android.content.res.ColorStateList;
-import android.database.Cursor;
 import android.graphics.Color;
 import android.net.Uri;
-import android.os.AsyncTask;
 import android.os.Bundle;
-import android.provider.OpenableColumns;
-import android.text.TextUtils;
 import android.util.Log;
 import android.view.Gravity;
 import android.view.Menu;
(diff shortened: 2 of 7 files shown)


thestinger:

Tommy-Geenexus:
@thestinger 
I had time to revisit and refactor/clean up the code.
I have noticed some bugs which remain unresolved in the app, and fixed them here in the process.

Also, i have revisited PDF specs 1.4 to 1.7 and made the date parser conform to those specs (they are a real PITA).

Tommy-Geenexus:
Thanks for the info, i'll update this pr to use support loaders and rebase once you've switched to appcompat.
Personally i think all mandatory features except search have been implemented already in my fork:

- Immersive mode

- Support encryption

- Option to rotate PDF's

Search will be by far the most challenging thing to implement, because traditional viewers use a recyclerview based approach, and im still not sure how to this here.
Touch zoom can be discarded in my opinion.

thestinger:
> Touch zoom can be discarded in my opinion.

I'd still like to have pinch-to-zoom. It can be implemented purely in JavaScript. It needs to scale using a CSS transform for speed and once the zoom completes it can render again and swap it in seamlessly.

thestinger:
@Tommy-Geenexus This is the approach for making zooming quick: https://github.com/CopperheadOS/platform_packages_apps_PdfViewer/commit/fef2ffda8a245efb7e29a4d022ba4bac19b1f5aa. For touch-based zoom, it would use this CSS style zooming until they release the pinch zoom. You can try this with a slow rendering PDF page and you'll see the zoom is near instant now and then it replaces with the proper zoomed render for better resolution when it's finished.

Tommy-Geenexus:
@thestinger 
Seems like a good approach, but the implementation should be done by
someone with more profound js/css experience, i'm afraid im not much of
a help in this compartment.

thestinger:
I made some other performance improvements too:

https://github.com/CopperheadOS/platform_packages_apps_PdfViewer/commit/cf7adb817578230d8b682b2c6d4a2524db25a895
https://github.com/CopperheadOS/platform_packages_apps_PdfViewer/commit/ae0bdbf8635b5146de5c3edc076af001ac0b7ebe

The prerendering could be a bit more clever by doing prerendering for someone jumping to a page number and moving backwards rather than forwards by tracking the direction of travel.

I also realized that a major performance issue is that Chromium canvas implementation on Android is ridiculously slow. It's a lot faster to render off-screen too, but showing the progress as it renders is nice... I'm not sure on the best approach to deal with that. It might take 100ms for a render to be mostly usable even if it takes 1-2s to finish and I don't want to hide it until it's done. I guess it could have a loop copying at intervals.

Tommy-Geenexus:
@thestinger 
I've updated this pr again, properties are now being displayed in a listview with a bold name,
which is how it should've been from the start.

Regarding the rendering, the app could show a spinner or something alike.

thestinger:
Thanks, I merged this but I left out the change to `TAG` for the time being and tweaked the commit message to say Overhaul instead of Refactor. I'll be looking it over some more and I'll make further changes in my own commits.

thestinger:
Down to 0 lint warnings now which is good.

thestinger:
Oh, and I did some more work on the performance overhaul. Needed to fix prerendering and also moved to the really awesome new streamTextContent API which drastically improves responsiveness for cancelling page rendering, since it's no longer blocking for a long time generating all the text content to send as one massive object.

PR #31, opened 2018-03-14
Loader issue workaround/fixes

Tommy-Geenexus:
- Fixes
- Update fragments to comply with the android developer guides.
- Supress remaining lint warnings

Also i've relocated the lifecycle methods in a natural order so that it's easier to analyze and read the code.

### DIFF
--- a/app/src/main/java/co/copperhead/pdfviewer/PdfViewer.java
+++ b/app/src/main/java/co/copperhead/pdfviewer/PdfViewer.java
@@ -33,14 +33,15 @@ import co.copperhead.pdfviewer.fragment.DocumentPropertiesFragment;
 import co.copperhead.pdfviewer.fragment.JumpToPageFragment;
 import co.copperhead.pdfviewer.loader.DocumentPropertiesLoader;
 
-public class PdfViewer extends Activity implements LoaderManager.LoaderCallbacks<List<CharSequence>> {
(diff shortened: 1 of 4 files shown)


app/src/main/java/co/copperhead/pdfviewer/PdfViewer.java
```diff
@@ -81,12 +85,8 @@ public void setNumPages(int numPages) {
 
         @JavascriptInterface
         public void setDocumentProperties(final String properties) {
-            if (mDocumentProperties != null) {
```
thestinger: I'd rather keep this sanity check, since we're meant to treat the JavaScript code as compromised here and there's no reason it should be calling this more than once that I can see. It makes a fresh instance for each document.

Tommy-Geenexus: Theoretically you are correct, but this throws an exception on each orientation change, due to the properties never being null (they are saved and restored in onCreate()).
Ideally this could be avoided, but due to a loader issue which is being triggered because the loader can't be initialized in onCreate(), the properties state must be saved, as the loader doesnt return the retrieved data after orientation change as it normally would.

There is an ugly workaround to this tough, if you want to keep the sanity check, which can be applied.

Tommy-Geenexus:
@thestinger 
I've reinserted the conditional check, and applied a workaround which goes more into depth.

PR #34, opened 2019-06-29
Don't hold static reference to fragment

Tommy-Geenexus:
Otherwise the activity won't be garbage collected and memory will be leaked.

### DIFF
--- a/app/src/main/java/org/grapheneos/pdfviewer/fragment/DocumentPropertiesFragment.java
+++ b/app/src/main/java/org/grapheneos/pdfviewer/fragment/DocumentPropertiesFragment.java
@@ -5,6 +5,7 @@ import android.app.Dialog;
 import android.os.Bundle;
 import android.widget.ArrayAdapter;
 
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/grapheneos/pdfviewer/fragment/DocumentPropertiesFragment.java
```diff
@@ -14,37 +15,35 @@
 import org.grapheneos.pdfviewer.R;
 
 public class DocumentPropertiesFragment extends DialogFragment {
-    public static final String TAG = "DocumentPropertiesFragment";
```
thestinger: Can you leave TAG as it was for now? I don't want to rely on `getSimpleName` without a clear understanding of how it will be impacted by proguard / r8.

Tommy-Geenexus: Sure.

PR #35, opened 2019-06-29
Add rotate document option

Tommy-Geenexus:
Ported to the new codebase with caching support.


Tommy-Geenexus:
@thestinger 
I'll port all reasonable leftover commits from the old codebase to the new one in the coming weeks.
Do you wan't me to replace the loader with a viewmodel or do you plan on sticking to loaders?

thestinger:
I'd like to get rid of loaders as long as the replacement doesn't make the code more complex.

app/src/main/java/org/grapheneos/pdfviewer/PdfViewer.java
```diff
@@ -282,6 +282,11 @@ private void renderPage(final boolean lazy) {
         mWebView.evaluateJavascript(lazy ? "onRenderPage(true)" : "onRenderPage(false)", null);
     }
 
+    private void documentOrientationChanged(final int orientationDegreesOffset) {
+        mWebView.evaluateJavascript(String.format("onDocumentOrientationChanged(\"%s\")",
```
thestinger: This should be a member variable on the Java side in order to save and restore the state, with it exposed via the Channel class like the page number and zoom level. I also want to avoid dynamically creating any JavaScript strings for various reasons.

Tommy-Geenexus: Done.

PR #39, opened 2019-07-01
Add skip to first/last page option

Tommy-Geenexus:
Ported to new codebase.

### DIFF
--- a/app/src/main/res/menu/pdf_viewer.xml
+++ b/app/src/main/res/menu/pdf_viewer.xml
@@ -38,6 +38,18 @@
         app:showAsAction="ifRoom" />
 
     <item
+        android:id="@+id/action_first"
+        android:icon="@drawable/baseline_first_page_24"
+        android:title="@string/action_first"
+        app:showAsAction="ifRoom" />
+
+    <item
+        android:id="@+id/action_last"
+        android:icon="@drawable/baseline_last_page_24"
+        android:title="@string/action_last"
+        app:showAsAction="ifRoom" />
+
+    <item
         android:id="@+id/action_jump_to_page"
         android:icon="@drawable/ic_pageview_white_24dp"
         android:title="@string/action_jump_to_page"
--- a/app/src/main/java/org/grapheneos/pdfviewer/PdfViewer.java
+++ b/app/src/main/java/org/grapheneos/pdfviewer/PdfViewer.java
@@ -310,8 +310,8 @@ public class PdfViewer extends AppCompatActivity implements LoaderManager.Loader
         }
     }
 
-    public void onJumpToPageInDocument(int selected_page) {
-        if (selected_page >= 1 && selected_page <= mNumPages) {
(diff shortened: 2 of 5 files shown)


app/src/main/res/menu/pdf_viewer.xml
```diff
@@ -37,6 +37,18 @@
         android:title="@string/action_zoom_in"
         app:showAsAction="ifRoom" />
 
+    <item
+        android:id="@+id/action_first"
+        android:icon="@drawable/ic_first_page_black_24dp"
```
thestinger: These should be the white icons rather than the black ones.

Tommy-Geenexus: Leftover from my old theme, now fixed.

Tommy-Geenexus:
@thestinger 
Before moving on to the more sophisticated stuff; Is it ok to make use of androidx libraries for e.g. dialogs?

thestinger:
@Tommy-Geenexus Yes, depending on any of the androidx libraries is fine. I originally didn't want to use the support libraries because it wasn't clear to me which direction Google was headed with it. It became clear that they expect every app to use it though, especially with deprecations of lots of non-core functionality in the OS APIs, and now androidx has reinforced that it's a set of API extensions.

I'd also be fine with eventually moving to Kotlin, but I'd rather keep using Java for now, and it would also make it a lot harder to recover your changes if it was migrated prematurely.

Tommy-Geenexus:
@thestinger 
Awesome, this enables more possibilities for improving the UI/UX.

I am not (yet) too familiar with Kotlin, although i'll learn it at some point (soon'ish).
But it would make sense to migrate sooner or later, as Kotlin comes with many benefits.

thestinger:
@Tommy-Geenexus https://github.com/GrapheneOS/PdfViewer/issues/30 is a good example of where I'd like to start using androidx.

thestinger:
@Tommy-Geenexus I'd really like to get encrypted PDF support landed by the way. I know you had a pull request for that before. There are a lot of things to do, but that's one of the basics.

Similarly, search would be really nice which mostly involves JavaScript code looking through the text layer and then highlighting the matching elements. I'm not quite sure how hard it will be yet. It would be fairly easy to do it within the current page, while searching the whole PDF will require more work since it will probably need to generate the text layers. Could have a look at how they do it in the official web viewer.

thestinger:
@Tommy-Geenexus Zoom gestures are another one of the basics, but it seems quite difficult to do it properly. I think the gestures should be detected in the JavaScript code rather than the outer app. It could reuse the existing code to handle zoomed rendering, but it would also need to center it where the user zoomed.

Tommy-Geenexus:
@thestinger 
Encrypted support is on the way, as is immersive mode and a few other minor improvements.
Search will require more time, as it needs deep integration into the action bar and the viewer.

Zoom gestures should prob have the highest priority, as they are essential.
I would prefer if someone with more in-depth js knowledge would take on that task.

PR #40, opened 2019-07-03
Add immersive mode

Tommy-Geenexus:


thestinger:
GitHub did something very weird with their built-in rebase support and I had to fix that... all good now.

thestinger:
GitHub did something very weird with their built-in 'rebase and merge' support and I had to fix that... all good now.

app/src/main/java/org/grapheneos/pdfviewer/PdfViewer.java
```diff
@@ -217,6 +219,31 @@ public void onPageFinished(WebView view, String url) {
             }
         });
 
+        final GestureDetector detector = new GestureDetector(PdfViewer.this,
+                new GestureDetector.SimpleOnGestureListener() {
+                    @Override
+                    public boolean onSingleTapUp(MotionEvent motionEvent) {
+                        if (mUri != null) {
+                            mWebView.evaluateJavascript("isTextSelected()", selection -> {
```
thestinger, line 227: Why does it need to check for text selection? Is it because it would lose the selection by doing this?

Tommy-Geenexus: This is to emulate the standard behaviour of other PDF viewers.
When text is selected, immersive mode should not be entered/exited,
the selection should be cleared beforehand.

thestinger:
It looks like this needs to be rebased on top of the port to pdf.js 2.x. The changes to viewer.js are minor so it won't be difficult.

Tommy-Geenexus:
Rebase done.

thestinger:
Thanks!

thestinger:
@Tommy-Geenexus Is it expected that the top of the PDF can't be seen without entering immersive mode? How does that work in other PDF viewers?

thestinger:
Ah, it shifts it up when first entering immersive mode which makes sense, but it doesn't shift it back to how it was before when exiting it.

thestinger:
@Tommy-Geenexus I filed https://github.com/GrapheneOS/PdfViewer/issues/44 about this. I think it should restore it back to how it was when the mode is exited. I'm not quite sure why it isn't already doing that though.

thestinger:
Alternatively, maybe this is a sensible way of doing it, but it just feels wrong to me that it switches the layout when first entering and it's not possible to get back to it.

thestinger:
Okay, I addressed this with https://github.com/GrapheneOS/PdfViewer/commit/4fad3956065cf3f15cc3ae1426a6df4bbbf2d4e4 which starts it off in that mode where the navigation covers the top and bottom of the rendered view. It would probably be a good idea to make the navigation transparent.

thestinger:
@Tommy-Geenexus So, ideally, it would have support for calculating the amount of space that's going to be taken up by the statusbar and navigation bar in order to put a black spacer at the top in the WebView, which is what the stock OS does in the PDF viewer. That way, the whole PDF is visible before the navigation bar is hidden, without shifting anything around, and with the whole screen used once the user starts scrolling down.

thestinger:
Filed https://github.com/GrapheneOS/PdfViewer/issues/45.

Tommy-Geenexus:
I'll take a look at it, although i'm busy for the next couple of weeks so it'll take some time.

thestinger:
@Tommy-Geenexus I implemented this with https://github.com/GrapheneOS/PdfViewer/commit/acfad2c09286c140412ed23e44df72bba0dcfade and it works how I want now. It starts off looking the same as before, but you can enter immersive mode without anything shifting around, and it simply has a reserved black area at the top.

thestinger:
@Tommy-Geenexus I implemented this with https://github.com/GrapheneOS/PdfViewer/commit/dbd0e3bb2dd7624778a120d04219ee8fce23810b which works almost exactly how I want now. It starts off looking the same as before adding immersive mode and you can enter immersive mode without anything shifting around. It simply has a reserved area at the top where the UI gets hidden as padding, so you can still use the whole screen, but the PDF view itself has some padding.

thestinger:
Still working on making it fully work.

thestinger:
https://github.com/GrapheneOS/PdfViewer/commit/be2439be6560e86501c82fa9a4c8dca96247182f seems to cover all the edge cases, but we'll see.

PR #46, opened 2019-10-05
Add gesture support

Virgiel:
I started to improve the gesture across the app by adding a pinch to zoom gesture.

As the gesture detectors and listeners stuff will become bigger and bigger, I wrote a helper class in another file that presents a simple API.

- [x] GestureHelper class
- [x] Pinch to zoom gesture
- [x] Edge swipe navigation

Deferred 
- Continuous zoom

This is my first time contributing to an open-source project, so be free to criticize and advise me.

### DIFF
--- a/app/src/main/assets/viewer.js
+++ b/app/src/main/assets/viewer.js
@@ -7,13 +7,12 @@ let renderPending = false;
 let renderPendingLazy = false;
 const canvas = document.getElementById('content');
 let orientationDegrees = 0;
-let zoomLevel = 100;
+let zoomRatio = 1;
 let textLayerDiv = document.getElementById("text");
-const zoomLevels = [50, 75, 100, 125, 150];
 let task = null;
 
 let newPageNumber = 0;
-let newZoomLevel = 0;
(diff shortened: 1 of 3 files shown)


thestinger:
The review feature is messing up, so I'll just post this as a comment:

It would be best to squash fa70f5e and 8e4d996 into the previous commit using git rebase -i PARENT_COMMIT and then changing this commit to squash instead of pick.

Virgiel:
To improve the user experience, I think we should use a continuous zoom with a temporary quick zoom between the new slow and lossless renderings. 
But this requires deeper changes, and I think that so far, this PR is already adding some useful gestures.

thestinger:
@Virgiel Thanks for all this work! I finally have time to merge it and do some follow-up work on it.

thestinger:
@Virgiel For some reason I can't get the swipe feature to work.

thestinger:
@Virgiel I'm leaving out the edge swipe gestures for now because it needs some work to make it more usable and reliable. It's possible that the leeway from the edges is just too strict.

PR #47, opened 2019-10-05
Add splash screen

Virgiel:
Show a splash screen while the app launch in a lightweight way.


thestinger:
I like the other proposals, but I'd rather not add a splash screen.

PR #48, opened 2019-10-05
Fix app bar menu item not being correctly enabled/disabled

Virgiel:


app/src/main/java/org/grapheneos/pdfviewer/PdfViewer.java
```diff
@@ -123,6 +123,7 @@ public int getDocumentOrientationDegrees() {
         @JavascriptInterface
         public void setNumPages(int numPages) {
             mNumPages = numPages;
+            invalidateOptionsMenu();
```
thestinger: I don't think it's correct to run this here since it's not running on the UI thread.

thestinger: This probably needs to use the function to run it on the main page, or it could just be done another way.

Virgiel: I fixed it with `runOnUiThread.`
I think it is preferable to call `invalidateOptionsMenu()` every time a value changes option menu state's, for the sake of simplicity.

thestinger:
I cherry-picked this to master manually since GitHub royally screwed it up.

PR #49, opened 2019-10-29
Disable launcher icon

YeQais:
because of this app will be hidden from app section when people need to open pdf file they can open it from clicking pdf file


Zoraver:
As a user of this app, I like having the launcher icon. In my opinion, if a user would like to hide the launcher icon, they should use a launcher with support for hiding specific apps.

thestinger:
I don't want to remove the launcher icon for this. I don't think normal untrusted apps should be able to avoid having a launcher icon at all.

PR #58, opened 2020-06-08
Add Polish translation

Atrate:
Signed-off-by: Atrate <Atrate@protonmail.com>


thestinger:
I cherry-picked the MIME change. We're not in a position to review or maintain translations to other languages right now. As a whole, our resources are stretched incredibly thin, many crucial things are being neglected and the things are barely being kept alive and ported forward.

I cannot review a translation for a language that I don't know, so I don't know if this is correct or not, and I will not be able to update it in the future. I cannot merge anything that I am unable to keep updated / maintained myself.

PR #59, opened 2020-06-08
Prettify README.md

Atrate:
Signed-off-by: Atrate <Atrate@protonmail.com>


thestinger:
I don't really want to change the paragraph into bullet points. Some of this reformatting is desirable, but I can't merge it as a whole. If you really want to do this I'd suggest doing it more incrementally so I don't have to reject it based on some of the changes even though other changes are desirable.

PR #61, opened 2020-07-08
Lossless optimize assets

tiagofilipesilva:


thestinger:
I need to be able to reproduce this myself. How are you optimizing these?

tiagofilipesilva:
Hi Daniel!

I used [optipng](http://optipng.sourceforge.net/) first for a quick check, then [Papa's Best Optimizer](https://papas-best.com/optimizer_en), which aggregates several lossless optimization tools. _Heads-up_: it takes a lot of time to process.

thestinger:
I don't really want to merge opaque changes.

thestinger:
I could do this myself but it's really not important / significant right now and I can't just merge changes to binary files.

tiagofilipesilva:
I actually agree with that, the part of minimizing binary files in software.
But the opacity would be the same, because one would also need to optimize the XML of the SVG files and would still be hard to read (better than binary, but still).
I know that you could do it, but you did not, because you don't have time. That's why I tried to help.
Maybe I could open an issue in GrapheneOS/PdfViewer so that someone with better vectorial skills than me could replace the bitmaps with vectorials?

PR #77, opened 2020-12-28
Print

sguerrini97:
I've added a menu option to print the document from the Uri.


sguerrini97:
Should close #36

louipc:
Thanks. Should rebase your branch to remove those rename/revert commits.

sguerrini97:
@louipc done. I should've done it before opening this PR...

Anyway, I was wondering: what do you guys think about feeding the untrusted PDF to the print service?
It is also possible to print the rendered webview but during my tests this broke the PDF layout (e.g. zooming on the webview would result in a zoom on the final print, ruining the whole PDF concept).

PR #79, opened 2021-01-03
bump gradle version to 4.1.1

anupritaisno1:
Signed-off-by: anupritaisno1 <www.anuprita804@gmail.com>


thestinger:
Should say Android gradle plugin instead of gradle, and it should be done for Auditor too.

PR #87, opened 2021-04-03
Add a menu item to share currently opened PDF

jan-de-haan:
I personally find it very useful to be able to share the currently opened file right from the PDF viewer itself, because I don’t have to remember where it came from (sometimes, the file might not even be saved anywhere easily accessible at all).

Sharing the currently opened file is implemented by providing the source URI to a sharing-intent. In the cases I’ve tried, other apps were always able to access this URI and share the file successfully.

Right now, the new menu item is the fourth one, after previous page, next page, and open document, but that just matches how often I personally use the menu items, so a different order might be better for most people.

The share icon itself is taken from [Google Material Icons](https://github.com/google/material-design-icons). They can be freely used and do not require any attribution.


emschu:
@jan-de-haan I just noticed that I did almost the same in PR #172 as you did here. The main difference is the handling of icons in your PR which are in `.png` format, although I think the icon assets should be in a vector-format (as in `res/drawable/ic_*.xml`). Would it be okay for you to close this PR in favor of PR #172 ?

jan-de-haan:
@emschu Yes, that’s okay, vector icons are definitely the better choice.

PR #88, opened 2021-04-21
Minor improvements in the UI/UX of PDF Viewer

MHShetty:
* Added grey background.
* Centered the viewport of the PDF viewer in the background.
* Default max zoom out when the document loads (to give the user an overview of the first page).


octocorvus (changes requested):
The 0.5f zoom value shouldn't be hardcoded.

PR #95, opened 2021-05-31
Polish translation

mkljczk:
Signed-off-by: marcin mikołajczak <me@mkljczk.pl>


thestinger:
We're not ready to maintain translations yet. It requires a long-term commitment to keeping them updated and we haven't worked out a system for this yet.

PR #105, opened 2021-08-11
Use grey background while rendering

MHShetty:


thestinger:
This is implemented now.

PR #112, opened 2021-08-30
automate pdf.js updates

flawedworld:


flawedworld:
this solves https://github.com/GrapheneOS/PdfViewer/issues/111

thestinger:
This will end up using the beta releases since they aren't distinguishing between them.

https://mozilla.github.io/pdf.js/getting_started/#download

PR #117, opened 2021-10-01
Enable full R8 optimisation

Zanthed:
Stays consistent with Auditor using full R8 optimisation.
https://github.com/GrapheneOS/Auditor/blob/main/gradle.properties


thestinger:
https://github.com/GrapheneOS/PdfViewer/commit/0d0bca6d4d9bfcbef734f26f121a5533884a35aa

PR #126, opened 2021-11-23
minor refactors

Zanthed:
Signed-off-by: June <zanthed@riseup.net>

### DIFF
--- a/app/src/main/java/org/grapheneos/pdfviewer/PdfViewer.java
+++ b/app/src/main/java/org/grapheneos/pdfviewer/PdfViewer.java
@@ -69,6 +69,7 @@ public class PdfViewer extends AppCompatActivity implements LoaderManager.Loader
         "document-domain=(), " +
         "encrypted-media=(), " +
         "fullscreen=(), " +
+        "gamepad=(), " +
         "geolocation=(), " +
         "gyroscope=(), " +
         "hid=(), " +
@@ -81,6 +82,7 @@ public class PdfViewer extends AppCompatActivity implements LoaderManager.Loader
         "picture-in-picture=(), " +
         "publickey-credentials-get=(), " +
         "screen-wake-lock=(), " +
+        "speaker-selection=(), " +
         "serial=(), " +
         "sync-xhr=(), " +
         "usb=(), " +
--- a/app/src/main/java/org/grapheneos/pdfviewer/Utils.java
+++ b/app/src/main/java/org/grapheneos/pdfviewer/Utils.java
@@ -39,22 +39,18 @@ public class Utils {
 
         final Calendar calendar = Calendar.getInstance();
         final int currentYear = calendar.get(Calendar.YEAR);
-        int year;
 
         // Year is required
         String field = date.substring(position += 2, 6);
         if (!TextUtils.isDigitsOnly(field)) {
             throw new ParseException("Invalid year", position);
         }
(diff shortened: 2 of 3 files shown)


app/src/main/java/org/grapheneos/pdfviewer/Utils.java
```diff
@@ -39,7 +39,7 @@ public static String parseDate(String date) throws ParseException {
 
         final Calendar calendar = Calendar.getInstance();
         final int currentYear = calendar.get(Calendar.YEAR);
-        int year;
+        int year = 0;
```
thestinger: It's normal to rely on zero initialization in Java.

app/src/main/java/org/grapheneos/pdfviewer/PdfViewer.java
```diff
@@ -213,6 +215,7 @@ public WebResourceResponse shouldInterceptRequest(WebView view, WebResourceReque
                     headers.put("Content-Security-Policy", CONTENT_SECURITY_POLICY);
                     headers.put("Permissions-Policy", PERMISSIONS_POLICY);
                     headers.put("X-Content-Type-Options", "nosniff");
+                    headers.put("Referrer-Policy", "no-referrer");
```
thestinger: I don't really see much point in adding this but we could do it if there's some reason for it.

app/src/main/java/org/grapheneos/pdfviewer/Utils.java
ghost: Always thought local variables weren't initialised automatically. Unless that's changed.

I guess it's more of a style thing since the other date variables are initialised.

Can be removed if you want.

app/src/main/java/org/grapheneos/pdfviewer/PdfViewer.java
ghost: Referrer-Policy isn't useful in this case for localhost so I guess maybe it also doesn't matter if it exists or not? The default is `strict-origin-when-cross-origin` so a referrer of somekind _is_ being sent. Whether it matters I'm not sure of.

Can be removed as I'm not aware of it providing any security benefits for a localhost PDF viewer.

app/src/main/java/org/grapheneos/pdfviewer/Utils.java
thestinger: For a local variable it just won't allow it to be used before it's initialized. If you want to clean this up you should just move the declaration where it's first initialized.

app/src/main/java/org/grapheneos/pdfviewer/PdfViewer.java
thestinger: Note it's not really using localhost but rather it just uses localhost as a placeholder and then intercepts all requests to localhost to provide them in the app code.

ghost: Ah okay. I knew there was something special when I saw `https://` infront of localhost.

I can remove Referrer-Policy as then there's nothing beneficial if there isn't an actual " web server".

thestinger: It has no network access so it can't actually access localhost. The method you're changing is the `shouldInterceptRequest` override which substitutes the requests to https://localhost/ with different content.

PR #131, opened 2021-12-27
minor refactorings and test cases

david-ruitter:


thestinger:
Seems to be failing the CI process right now.

david-ruitter:
@thestinger Seems like the decimal tests are failing in the Pipeline, but it does not give further Information what is going wrong. They are running on my machine if I use the same gradle command. Do you have an idea what's the problem here?

image

thestinger:
Needs to be rebased and whatever is wrong with the tests need to be fixed.

PR #132, opened 2021-12-30
update pdf.js to 2.12.313

Zanthed:


Zanthed:
We want to figure out how to automate updating pdf.js instead.

Zanthed:
Implemented in https://github.com/GrapheneOS/PdfViewer/pull/138

PR #134, opened 2022-01-14
Add Polish translation

temp-byte:


thestinger:
We're not ready to manage translations yet. This is wanted but we need to put together translation teams for each language we intend to do this for so that it gets reviewed properly, maintained and updated whenever the strings change. It's not something we can accept as is.

PR #143, opened 2022-02-06
Solve some PdfViewer.java deprecations and warnings

Zanthed:


app/src/main/java/org/grapheneos/pdfviewer/PdfViewer.java
```diff
@@ -218,7 +218,7 @@ public WebResourceResponse shouldInterceptRequest(WebView view, WebResourceReque
                     headers.put("Content-Security-Policy", CONTENT_SECURITY_POLICY);
                     headers.put("Permissions-Policy", PERMISSIONS_POLICY);
                     headers.put("X-Content-Type-Options", "nosniff");
-                    response.setResponseHeaders(headers);
+                    Objects.requireNonNull(response).setResponseHeaders(headers);
```
thestinger: I would rather not use this approach right now.

ghost: Removed Object.requireNonNull

PR #146, opened 2022-02-07
Fix usage of deprecated getSystemWidnowInsets

jann-0:


Zanthed:
I think we'd actually want to replace insets all together and use a proper layout like so here (not finished): https://github.com/smdyv/PdfViewer/commit/267ff843e6876390d6391f8ae68a6fda104b742e

jann-0:
Makes sense.

Zanthed:
https://github.com/GrapheneOS/PdfViewer/pull/154

jann-0:
Thanks for that follow up!

PR #175, opened 2022-03-10
Add italian translation

unbranched:


unbranched:
Ok, just read the message in the chinese pull request... well, will be waiting.

PR #185, opened 2022-04-14
Add Polish translation

PatrykMis:
Since it is not app for use with only GrapheneOS and it is available on Google Play, I thought it's time for adding translations. My friends use it successfully e.g. on Xiaomi/Samsung phones/tablets with LineageOS.


thestinger:
This needs to wait until the completion of moving strings to resources and forming official translation teams.

PR #192, opened 2022-04-28
RU translation

maade69:


thestinger:
This needs to wait until the completion of moving strings to resources and forming official translation teams.

PR #199, opened 2022-05-12
Ui improvements 

empratyush:
- Use dynamic color on supported devices
- fix navigation bar color issue on three button navigation setting (on api 28). [before](https://i.postimg.cc/brQvK2Mx/before.png) vs [after](https://i.postimg.cc/HLcdqy8r/after.png)

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -5,6 +5,7 @@
     <original-package android:name="org.grapheneos.pdfviewer" />
 
     <application
+        android:name=".App"
         android:allowBackup="true"
         android:icon="@mipmap/ic_launcher"
         android:label="@string/app_name"
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -13,6 +13,7 @@
         android:theme="@style/AppTheme">
         <activity
             android:name=".PdfViewer"
+            android:documentLaunchMode="always"
             android:exported="true">
             <intent-filter>
                 <action android:name="android.intent.action.MAIN" />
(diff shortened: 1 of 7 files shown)


app/src/main/AndroidManifest.xml
```diff
@@ -15,6 +15,7 @@
         android:theme="@style/AppTheme">
         <activity
             android:name=".PdfViewer"
+            android:launchMode="singleTop"
```
muhomorr: I think declaring `android:documentLaunchMode="always"` instead of changing `launchMode` would be better.
https://developer.android.com/guide/topics/manifest/activity-element#dlmode

Then, OS will create a new task with a PdfViewer activity in it each time PDF is opened, which means that each PDF will be shown as a separate item in the task switcher UI.

Moreover, this will allow to open the same document multiple times.

empratyush: It can create too many task instance in task switcher UI / recent screen. tho unsure if that's what we want.

thestinger: I think it's probably what we want.

thestinger: Document mode is probably what we want.

empratyush: okay, thanks.

app/src/main/java/app/grapheneos/pdfviewer/PdfViewer.java
```diff
@@ -393,6 +390,22 @@ public void onZoomEnd() {
         }
     }
 
+    @Override
+    protected void onNewIntent(Intent intent) {
```
muhomorr: `onNewIntent` won't ever be called when `android:documentLaunchMode="always"` is set.
A new instance of `PdfViewer` activity will be created each time.

empratyush: okay,

thestinger:
@empratyush Needs to be rebased due to conflict in `app/src/main/AndroidManifest.xml`.

PR #213, opened 2022-06-15
Vertically and horizontally center PDF page

MHShetty:
A custom implementation for supporting zooming and moving across the document has been added (in commit f7b9ae7ea651cf3fa201d6001d5fe60ad73f399f) to support the changes related to centering the document.

Without the custom implementation, the user won't be able to properly move across the document, specifically the top-left corner (as the user gets restricted to the actual bounds of the page [once they have sufficiently zoomed in]).

We might require to test or review this PR before we merge these changes.

### DIFF
--- a/app/src/main/assets/viewer.css
+++ b/app/src/main/assets/viewer.css
@@ -6,7 +6,7 @@ body, canvas {
 canvas {
     margin: auto;
     display: block;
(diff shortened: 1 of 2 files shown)


thestinger:
Doing this via #229.

PR #229, opened 2022-11-20
horizontally and vertically center canvas and text layer

octocorvus:
Also fixes an issue where text layer is not exactly on top of the canvas when cavas width is less than screen width (which happens often in landscape mode or when fully zoomed out)

### DIFF
--- a/app/src/main/assets/viewer.css
+++ b/app/src/main/assets/viewer.css
@@ -1,24 +1,36 @@
+html, body {
+    height: 100%;
+}
+
 body, canvas {
     padding: 0;
     margin: 0;
-}
-
-canvas {
-    margin: auto;
-    display: block;
 }
 
 body {
     background-color: #c0c0c0;
 }
 
+#container {
+    width: 100%;
+    height: 100%;
+    display: grid;
+    place-items: center;
+}
+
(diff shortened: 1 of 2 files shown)


octocorvus:
since my changes rely purely on html/css and contains way less code, i think it is also an overall better implementation of what #213 tries to do, so merging my changes should also close #213

as far as testing goes, i've tested my changes in both landscape and portrait mode with pdf of various shapes and sizes and so far there doesn't appear to be any regressions and layout appears as expected in all of these four cases (at most four cases are possible depending on the zoom value and pdf's width-height ratio):
* pdf width < screen width && pdf height < screen height: pdf is centered both horizontally and vertically
* pdf width < screen width && pdf height >= screen height: pdf is centered horizontally and occupies full height
* pdf width >= screen width && pdf height < screen height: pdf occupies full width and is centered vertically
* pdf width >= screen width && pdf height >= screen height: pdf occupies full width and height

by 'pdf occupies full length (width or height)', i mean that the pdf covers the full length of the screen and the hidden part can be made visible by scrolling in the appropriate direction

bugs related to zooming in and out (e.g. pdf goes out of the view and jumps back in some cases) remain, which i'm planning to fix via another pull request

PR #267, opened 2023-05-31
set zoomRatio to fit document to screen size

Hanouta:
This adjusts the zoom ratio after opening a document (or rotating) so the document fits the whole screen.
It also sets the minimum zoom ratio to the same value, because there is not really a need to zoom out further.

Closes: #184, Closes: #201

### DIFF
--- a/app/src/main/assets/viewer.js
+++ b/app/src/main/assets/viewer.js
@@ -72,6 +72,13 @@ function setLayerTransform(pageWidth, pageHeight, layerDiv) {
     layerDiv.style.translate = `${translate.X}px ${translate.Y}px`;
 }
 
+function getDefaultZoomRatio(page, orientationDegrees) {
+    const viewport = page.getViewport({scale: 1, rotation: orientationDegrees});
+    const widthZoomRatio = document.body.clientWidth / viewport.width;
+    const heightZoomRatio = document.body.clientHeight / viewport.height;
+    return Math.max(Math.min(widthZoomRatio, heightZoomRatio, channel.getMaxZoomRatio()), channel.getMinZoomRatio());
+}
+
 function renderPage(pageNumber, zoom, prerender, prerenderTrigger=0) {
     pageRendering = true;
     useRender = !prerender;
@@ -106,6 +113,14 @@ function renderPage(pageNumber, zoom, prerender, prerenderTrigger=0) {
     pdfDoc.getPage(pageNumber).then(function(page) {
         if (maybeRenderNextPage()) {
             return;
+        }
+
+        const defaultZoomRatio = getDefaultZoomRatio(page, orientationDegrees);
+
+        if (cache.length === 0) {
+            zoomRatio = defaultZoomRatio;
+            newZoomRatio = defaultZoomRatio;
+            channel.setZoomRatio(defaultZoomRatio);
         }
 
         const viewport = page.getViewport({scale: newZoomRatio, rotation: orientationDegrees});
--- a/app/src/main/java/app/grapheneos/pdfviewer/PdfViewer.java
+++ b/app/src/main/java/app/grapheneos/pdfviewer/PdfViewer.java
@@ -108,7 +108,7 @@ public class PdfViewer extends AppCompatActivity implements LoaderManager.Loader
         "usb=(), " +
         "xr-spatial-tracking=()";
 
-    private static final float MIN_ZOOM_RATIO = 0.5f;
+    private static final float MIN_ZOOM_RATIO = 0.2f;
     private static final float MAX_ZOOM_RATIO = 1.5f;
     private static final int ALPHA_LOW = 130;
     private static final int ALPHA_HIGH = 255;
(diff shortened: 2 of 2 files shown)


app/src/main/assets/viewer.js
```diff
@@ -108,6 +116,18 @@ function renderPage(pageNumber, zoom, prerender, prerenderTrigger=0) {
             return;
         }
 
+        const defaultZoomRatio = getDefaultZoomRatio(page, orientationDegrees);
+
+        if (resetZoomRatio) {
+            zoomRatio = defaultZoomRatio
+            newZoomRatio = defaultZoomRatio;
+            channel.setZoomRatio(defaultZoomRatio);
```
octocorvus: I think, zoom ratio should only be reset on first load, not subsequent orientation changes (which is the current behaviour). However, if zoom ratio is less than the default zoom ratio (current minimum), then zoom ratio should be reset.

app/src/main/assets/viewer.js
```diff
@@ -108,6 +116,18 @@ function renderPage(pageNumber, zoom, prerender, prerenderTrigger=0) {
             return;
         }
 
+        const defaultZoomRatio = getDefaultZoomRatio(page, orientationDegrees);
+
+        if (resetZoomRatio) {
+            zoomRatio = defaultZoomRatio
```
octocorvus: Add missing semicolon.

app/src/main/assets/viewer.js
```diff
@@ -108,6 +116,18 @@ function renderPage(pageNumber, zoom, prerender, prerenderTrigger=0) {
...
+            newZoomRatio = defaultZoomRatio;
+            channel.setZoomRatio(defaultZoomRatio);
+            channel.setMinZoomRatio(defaultZoomRatio);
```
octocorvus: Setting minimum zoom ratio is problematic with PDFs of different page sizes. PDF Viewer prerenders the next and previous pages; and if those pages have different size than the current page, the minimum zoom ratio will be set to an incorrect value for the current page. Test using [sizes.pdf](https://github.com/mozilla/pdf.js/blob/master/test/pdfs/sizes.pdf).

octocorvus (changes requested):
~Also, this does not reset minimum zoom ratio when fullscreen mode is toggled (using single tap). Again, I think, the actual zoom ratio should only be reset if it is less than the default zoom ratio (current minimum).~

app/src/main/assets/viewer.js
Hanouta: Added

app/src/main/assets/viewer.js
Hanouta: I thought of this possibility, thats why the minimum zoom ratio is being lowered if needed: https://github.com/GrapheneOS/PdfViewer/blob/b49b48497384f53e997b21c566bdfe044e6d12c5/app/src/main/assets/viewer.js#L127-L129

This way the user can always zoom out far enough to see the whole page. Ofc it would be possible to reset the minimum zoom ratio after switching to a different page, so the user can't zoom out "too far" on some pages, but I'm not sure if this is even worth implementing, especially when thinking of the possibility of a "vertical scrolling" feature in the future.

octocorvus: The main issue is inconsistency, on some PDF pages you cannot zoom out "too far", but on some you can. I suggest removing `minZoomRatio`, and adding back the constant `MIN_ZOOM_RATIO` (also lower its value), so that user experience is consistent.

Hanouta: I don't think there is a perfect solution to this, there are 3 possibilities:

- setting a fixed MIN_ZOOM_RATIO:
  - most documents will have a border when zoomed out fully
  - in case of different page sizes, the border won't be the same size on every page
  - when opening a pdf with a HUGE page size, it won't even be possible to zoom out far enough to see the whole document at once (unless setting MIN_ZOOM_RATIO to something ridiculous like 0.1, which could cause confusion on regular sized pages)
  - when opening a pdf with small page sizes, it would be possible to zoom out to a point where the page would just be barely visible on the screen
- setting minZoomRatio based on the smallest page:
  - won't have borders if all pages are the same size
  - borders on some pages if page sizes are not the same
- adjusting minZoomRatio based on the currently visible page:
  - it would likely not be possible to combine that with vertical scrolling, which is the most upvoted issue in the issue tracker here by far
  - otherwise no obvious downsides, should be consistent on every page, no matter the size

Another alternative would be variant 2, but setting minZoomRatio to something like `defaultZoomRatio * 0.80`, this way all pages would have a border when zoomed out fully, just differently sized, just like variant 1 without some of its downsides.

Personally I think the best approach long-term would be the alternative approach to variant 2, where the minZoomRatio would be slightly lower than the defaultZoomRatio. Let me know what you think

octocorvus: The first approach has a bunch of advantages:
* Simplest logic from all three (as there is no need to set minimum zoom ratio).
* When vertical scrolling is implemented, it would be nice to be able to zoom out further to have an overview of a bunch of pages at once. So, I don't think borders when fully zoomed out are an issue.

As for opening pages with ridiculously large (or small) page sizes, I think those are rare edge cases. A value of around 0.2 (20%) should work fine for almost all PDFs.

The second approach is inconsistent, and slightly more complex than first. The third approach might get too complex, and likely incompatible with vertical scrolling (like you said above). So, I think it is best to go with first approach, for now.

Hanouta: Changed it back to the static MIN_ZOOM_RATIO, and set it to 0.2

app/src/main/assets/viewer.js
Hanouta: From a consistency standpoint it would (imo) make sense to reset after hitting the "rotate clockwise/counterclockwise" button, because when actually rotating the whole device it would be reset too because the webview gets recreated.

octocorvus: It makes sense to preserve zoom ratio (set by the user). Also, webview shouldn't need to reload after device is rotated, it is a different issue.

Hanouta: Ah, I just assumed the webview reload was intentional (maybe it would add a lot of complexity to handle the whole change or something)

I removed the part which reset the zoom ratio after changing orientation, and I also removed the `resetZoomRatio` variable because the first document load can also be detected by checking if the cache is empty.

octocorvus:
Thank you for the pull request! As a final change, please squash all the commits.

octocorvus (approved):

PR #272, opened 2023-06-26
Typescript support

octocorvus:


macie:
Could you add more details about this change? What is wrong with plain JS scripts? Are there any risks in introducing more _npm_ dependencies (with/without caret ranges)?

In other words: what are advantages and disadvantages of your change from security and maintenance standpoints?

thestinger:

PR #278, opened 2023-07-04
Opt in to predictive back gesture

soupslurpr:
https://developer.android.com/guide/navigation/custom-back/predictive-back-gesture


thestinger:

soupslurpr:
done

PR #282, opened 2023-07-15
add navigating back and forth in PDF by swiping right or left

heapifyman:
Intends to implement #41 

The threshold values are a bit arbitrary but seem to be working fine on a Pixel 6 with latest GrapheneOS.

Swiping from the left or right edge will still "close" the app as it does in current version [17](https://github.com/GrapheneOS/PdfViewer/releases/tag/17).


thestinger:

mrcbax:
What is going on with this circus? Force pushing should not be abused in this way.

What is the hold up on merging this? It should have made it in the v18 release. This is a top requested community feature. Has been since issue #41 all the way to issue #202 

Stop blaming the community for slowing you down and tell us what needs to be done instead of one many army force pushing all over the codebase.

PR #319, opened 2023-12-18
Render below status bar in portrait orientation on devices with a display cutout

soupslurpr:
On devices with a display cutout, instead of hiding the status bar which shows a black bar, this makes it render below the bar by not hiding it when in portrait. Also uses the newer way to get the WindowInsetsController for SDKs supporting it.


soupslurpr:
Yeah sure, here you go:

Before:
Screenshot_20231218_120056


After:
Screenshot_20231218_115954

soupslurpr:
Target SDK 35 makes apps render edge-to-edge. Version 20 renders it fullscreen without the status bar being visible

PR #383, opened 2024-08-15
Added in swipe gestures to navigate pages

Madmatt88:
Address issue #41 and adds the feature of being able to use swipe gestures to navigate the pages of a PDF.


Madmatt88:
While this is still pending, I will continue to test this build and fix any other issues with the implementation.

Madmatt88:
Yes, when I started this I didn’t realise that there was another PR doing the same thing. I would like to see this merged but I haven’t heard anything more on it. I think it needs some work to get it up to scratch with the rest of the code and would like that help.

Madmatt88:
@thestinger Thank you for that piece of information, I will have a look and change the implementation to fit that requirement. 

@alt3r-3go  I will rebase the PR and re-trigger the action to find the build error. After that I'll start to make the changes to fit the new requirements and then update the PR, if I have any major issues I will reach out.

app/src/main/java/app/grapheneos/pdfviewer/GestureHelper.java
```diff
@@ -24,9 +32,20 @@ static void attach(Context context, View gestureView, GestureListener listener)
         final GestureDetector detector = new GestureDetector(context,
                 new GestureDetector.SimpleOnGestureListener() {
                     @Override
-                    public boolean onSingleTapUp(MotionEvent motionEvent) {
+                    public boolean onSingleTapUp(@NonNull MotionEvent motionEvent) {
```
alt3r-3go, line 35: This seems unrelated to the PR topic, I'd suggest removing, to avoid polluting git history.

app/src/main/java/app/grapheneos/pdfviewer/GestureHelper.java
```diff
@@ -7,13 +7,21 @@
...
     The GestureHelper present a simple gesture api for the PdfViewer
 */
 
 class GestureHelper {
     public interface GestureListener {
         boolean onTapUp();
+        boolean onFling(@Nullable MotionEvent e1, @NonNull MotionEvent e2, float velocityX, float velocityY);
+        boolean onDown();
+        // Can be replaced with ratio when supported
+        void onZoomIn(float value);
```
alt3r-3go, line 23: I wonder why would this and the next one are needed, given it's not about zooming?

app/src/main/java/app/grapheneos/pdfviewer/PdfViewer.java
```diff
@@ -385,8 +387,41 @@ public boolean onTapUp() {
...
+                    public void onZoomIn(float value) {
+                        zoomIn(value, false);
+                    }
+
+                    @Override
+                    public void onZoomOut(float value) {
+                        zoomOut(value, false);
+
                     public void onZoom(float scaleFactor, float focusX, float focusY) {
-                        zoom(scaleFactor, focusX, focusY, false);
```
alt3r-3go: Whitespace-only changes like this are usually discouraged, or at least should be split out to dedicated commits.

app/src/main/java/app/grapheneos/pdfviewer/PdfViewer.java
```diff
@@ -385,8 +387,41 @@ public boolean onTapUp() {
...
                     @Override
+
+                    public boolean onFling(@Nullable MotionEvent e1, @NonNull MotionEvent e2, float velocityX, float velocityY) {
+                        assert e1 != null;
+                        float maxXVelocity = 4000;
+                        float deltaX = e2.getX() - e1.getX();
+                        float deltaY = e2.getY() - e1.getY();
+
+                        if (Math.abs(deltaX) > Math.abs(deltaY)){
+                            if ((deltaX > 0) & (Math.abs(velocityX) > maxXVelocity)) {
```
alt3r-3go, line 465: Bitwise & operator used instead of the boolean one.

app/src/main/java/app/grapheneos/pdfviewer/PdfViewer.java
```diff
@@ -385,8 +387,41 @@ public boolean onTapUp() {
...
+                                Log.d("Horizontal", "Right movement Position");
+                                onJumpToPageInDocument(mPage - 1);
+                            } else if ((Math.abs(velocityX) > maxXVelocity)) {
```
alt3r-3go, line 468: I'd suggest moving this condition over to the higher-level `if` clause, to simplify and avoid repeating it.

app/src/main/java/app/grapheneos/pdfviewer/PdfViewer.java
```diff
@@ -385,8 +387,41 @@ public boolean onTapUp() {
...
+                                Log.d("Horizontal", "Left movement Position");
+                                onJumpToPageInDocument(mPage + 1);
+                            }
+                        }
+                        return true;
+                    }
+
+                    @Override
+                    public boolean onDown() {
```
alt3r-3go, line 477: Is this one really necessary?

alt3r-3go (commented):
Added a few comments, hopefully helpful. I'd also suggest taking a look at the other PR on this, as it seems to have pieces worth merging here.

Madmatt88:
@alt3r-3go  thanks for those comments and one thing that will help this is decoupling the page change from the onFling system call. I had to add in extra conditions in there because I had big problems with the page changing while i was reading as I was moving the page from left to right. Some things I did I can’t remember fully because it has been so long ago.

Madmatt88:
@alt3r-3go  I have re-run the action and it gave these errors:

``` bash
> Task :app:compileDebugKotlin FAILED
e: file:///home/runner/work/PdfViewer/PdfViewer/app/src/main/java/app/grapheneos/pdfviewer/fragment/JumpToPageFragment.kt:49:32 Unresolved reference 'onJumpToPageInDocument'.
e: file:///home/runner/work/PdfViewer/PdfViewer/app/src/main/java/app/grapheneos/pdfviewer/fragment/PasswordPromptFragment.kt:85:37 Unresolved reference 'loadPdfWithPassword'.
```
I am not quite sure why it can't find this reference, I have run the gradle command locally and it fails in the same section and with extra compiler flags provided other errors that didn't make sense the fix the problem. The areas that are now causing issues also were never changed so I have no idea what could cause this one. There should be no reason why it can't find the reference since the import seems correct on the files as well. 

Have you seen this before? 

Thank you for your help.

Madmatt88:
So I have made one fatal mistake by not creating a new feature branch which has created a tangled web of commits and changes causing my fork to cause problems. For future work I will not be making that mistake again. 

I still want to keep contributing to this project and have learnt a lot by going through this PR process, if @alt3r-3go you are able to get the project to build and make the changes with ease then I am more than happy for you to that. 

The other option is that I will start again locally, close this PR and submit a new one that doesn't have the same headaches so it can still be added. 

Thanks everyone for your patience.

Madmatt88:
With #603 doing this feature, I don’t see the need to keep this PR open as it just adds noise.

PR #432, opened 2025-01-01
Replace interest-cohort with browsing-topics

colleirose:
`interest-cohort` is no longer recognized by any browsers, and has been superseded by `browsing-topics`. See https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Permissions-Policy/browsing-topics and https://www.theverge.com/2022/1/25/22900567/google-floc-abandon-topics-api-cookies-tracking

### DIFF
--- a/app/src/main/java/app/grapheneos/pdfviewer/PdfViewer.java
+++ b/app/src/main/java/app/grapheneos/pdfviewer/PdfViewer.java
@@ -81,7 +81,8 @@ public class PdfViewer extends AppCompatActivity implements LoaderManager.Loader
         "accelerometer=(), " +
         "ambient-light-sensor=(), " +
         "autoplay=(), " +
-        "battery=(), " +
+        "battery=(), " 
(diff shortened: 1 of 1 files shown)


colleirose:
I see - I will try to create a permissions policy that matches these features.

colleirose:
> We probably don't want to add all of them and some are for enabling rather than disabling features. For now, just want to focus on whether we should actually remove interest-cohort. It only matters whether it's **actually** deprecated in Chromium not what others say about it with no say in that.

My bad, I see. Could you send a link to where you can find the list of what permissions are supported in Chromium?

colleirose:
According to the articles I've read about this, they silently disabled FLoC without an announcement in 2021, then announced the Topics API as a successor in 2022. So it's possible the permission still exists but is just not being used. I will check the Chromium source code and see if I can find anything, though I might need to head to sleep in a few hours.

PR #511, opened 2025-06-28
feat: add vertical scrolling mode for PDF viewer

cameronaaron:
feat: add vertical scrolling mode for PDF viewer
- Implemented vertical scrolling mode to display all PDF pages in a continuous scroll view.
- Added styles for vertical layout and page number indicators in CSS.
- Enhanced JavaScript logic for rendering all pages and tracking current page based on scroll position.
- Introduced a toggle button in the toolbar for switching between vertical scroll mode and traditional page mode.
- Updated Android integration to support vertical scroll mode and preserve state during mode switching.
- Added new vector drawable for the vertical scroll toggle icon.


viewer/js/index.js
```diff
@@ -28,11 +28,22 @@ const maxCached = 6;
 
 let isTextLayerVisible = false;
 
+// Vertical scrolling mode variables
+let isVerticalScrollMode = false;
```
ggtlvkma356, line 32: Better not to have two sets of states. Better use the state from Java side.

cameronaaron:
> There is a bug when the pdf file is hundreds of pages long, WebView will run out of vertical scrolling space and you cannot scroll to the beginning of the pdf.

Oh and you hit that?

viewer/js/index.js
```diff
@@ -86,6 +97,257 @@ function getDefaultZoomRatio(page, orientationDegrees) {
     return Math.max(Math.min(widthZoomRatio, heightZoomRatio, channel.getMaxZoomRatio()), channel.getMinZoomRatio());
 }
 
+// Vertical scrolling mode functions
+function createAllPagesContainer() {
```
ggtlvkma356, line 101: IMHO if we are going to create the canvas on the fly, then it is better to make both modes dynamically created, rather than having mixed ways of showing them.

gradle/verification-metadata.xml
ggtlvkma356, line 1: No need to add `sha256` checksums, `sha512` is fine.

app/src/main/res/drawable/ic_view_agenda_24dp.xml
ggtlvkma356, line 1: The tint color looks wrong to me.

rdevshp:
@cameronaaron Are you still working on this PR?

PR #570, opened 2025-12-01
Pdfjs update 5.4.449

octocorvus:


inthewaves:
pdf.js has already been updated to a more recent version

PR #573, opened 2025-12-07
Improved jump to page UI to be more intuitive to use

xpncvr:
The user immediately knows to type the page number instead of having to use a scroll wheel. 

Other people found it unintuitive too: https://github.com/GrapheneOS/PdfViewer/issues/236

It looks like this
image

### DIFF
--- a/app/src/main/java/app/grapheneos/pdfviewer/fragment/JumpToPageFragment.kt
+++ b/app/src/main/java/app/grapheneos/pdfviewer/fragment/JumpToPageFragment.kt
@@ -1,60 +1,56 @@
 package app.grapheneos.pdfviewer.fragment
 
 import android.app.Dialog
-import android.content.DialogInterface
 import android.os.Bundle
+import android.text.InputType
 import android.view.Gravity
+import android.widget.EditText
 import android.widget.FrameLayout
-import android.widget.NumberPicker
 import androidx.fragment.app.DialogFragment
 import app.grapheneos.pdfviewer.PdfViewer
+import app.grapheneos.pdfviewer.R
 import com.google.android.material.dialog.MaterialAlertDialogBuilder
 
 class JumpToPageFragment : DialogFragment() {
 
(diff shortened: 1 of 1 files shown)


app/src/main/java/app/grapheneos/pdfviewer/fragment/JumpToPageFragment.kt
```diff
@@ -1,60 +1,56 @@
...
-            mPicker.maxValue = viewerActivity.mNumPages
-            mPicker.value = viewerActivity.mPage
-        }
-        val layout = FrameLayout(requireActivity())
-        layout.addView(
-            mPicker, FrameLayout.LayoutParams(
-                FrameLayout.LayoutParams.WRAP_CONTENT,
+        val viewerActivity = requireActivity() as PdfViewer
+
+        val input = EditText(requireContext()).apply {
```
ggtlvkma356, line 23: The layout can be specified using `xml`.

app/src/main/java/app/grapheneos/pdfviewer/fragment/JumpToPageFragment.kt
```diff
@@ -1,60 +1,56 @@
...
-        } else {
-            mPicker.minValue = 1
-            mPicker.maxValue = viewerActivity.mNumPages
-            mPicker.value = viewerActivity.mPage
-        }
-        val layout = FrameLayout(requireActivity())
-        layout.addView(
-            mPicker, FrameLayout.LayoutParams(
-                FrameLayout.LayoutParams.WRAP_CONTENT,
+        val viewerActivity = requireActivity() as PdfViewer
```
ggtlvkma356, line 21: The old code uses a cast but I think it is better to pass whatever you need as arguments to the fragment.

inthewaves:
Handled by https://github.com/GrapheneOS/PdfViewer/pull/684

PR #579, opened 2025-12-26
Add find in document (fixes #4)

### LINKED ISSUE #4: search

JuliusBairaktaris:
### Description

Adds find-in-document to the PDF viewer. Fixes #4.

This is a full rewrite of the original version of this PR. That one was written against the
pre-Kotlin `PdfViewer.java` tree and no longer applies; more importantly it put the search engine
in JavaScript (a `SearchController` class holding all the state) and hand-rolled the parts the
platform already provides. This version inverts that: **Kotlin owns the search engine, JavaScript
is reduced to a text pump and a `Range` factory**, and both the matcher and the highlight renderer
are platform built-ins rather than new code.

### Architecture

| Concern | Where | What does the work |
| --- | --- | --- |
| Match index, counts, navigation, cancellation | Kotlin (`search/DocumentSearch.kt`) | `TreeMap` of page → matches |
| Text matching | Kotlin | **`android.icu.text.StringSearch`** at `Collator.PRIMARY` |
| Text extraction | JS (`viewer/js/search.js`, `index.js`) | `page.getTextContent()` |
| Offset → DOM mapping | JS | `TextLayer.textDivs` (pdf.js public API) |
| Highlight painting, scaling, rotation | Browser | **CSS Custom Highlight API** |
| UI | Kotlin (`SearchAppBar`) | Material 3 `TopAppBar` |

**Why ICU.** `StringSearch` is the same collation search engine Chromium's own find-in-page uses.
It folds case, diacritics, ligatures, full-width forms, ß/ss and CJK, and it reports match offsets
**into the original string**. That last property is what removes the need for a normalized copy of
the page text plus a map back to it — the roughly 900 lines of `normalize()` + diff-array machinery
that pdf.js's own `PDFFindController` needs. Nothing is vendored and no dependency is added.

**Why the Custom Highlight API.** It paints from live `Range` objects, so highlights follow the
text layer through zoom, rotation and the canvas insets with no coordinate arithmetic anywhere —
no `getClientRects()`, no dividing by `--scale-factor`, no overlay `<div>`s. It also mutates no
DOM, which matters because `index.js` keeps an LRU cache of six rendered text layers and reuses
those exact nodes; the DOM-surgery approach pdf.js uses would corrupt them on every cache hit.

**Bundle cost: +1.5 KB minified.** For comparison, importing `PDFFindController` from
`pdfjs-dist/web/pdf_viewer.mjs` costs **+164 KB** and needs a `globalThis.pdfjsLib` shim, because
that file destructures ~60 names off it at module scope. That option was measured and rejected.

### Bridge

One new `@JavascriptInterface` method, and two calls the other way:

- `setPageText(page: Int, itemsJson: String, generation: Int): Boolean` — JS ships one page of
  per-item strings. The Boolean return is the entire cancellation protocol: `false` means the index
  is full, stop. `generation` fences a sweep started for a previously opened document.
- `extractText(startPage, generation)` — walks outward from the page being viewed so the first result lands
  immediately on a large document, then wraps.
- `setSearchHighlights(page, [[[itemIndex, offset, length], …], …], active)` — digits, commas and
  brackets only.

**The query string never crosses the bridge in either direction**, so there is nothing to escape.

### UI

Follows the find-in-page pattern of Chrome, Firefox and Acrobat: a magnifier in the toolbar opens
a bar that replaces the top app bar — back arrow, text field, clear button, match counter, and
previous/next. Search-as-you-type with a 200 ms debounce, a progress bar while the document is
still being indexed, back button closes, IME "Search" jumps to the next match and dismisses the
keyboard.

| Find bar | First match | Stepping matches | Zoomed to 180% |
| --- | --- | --- | --- |
|  |  |  |  |

Highlights are magenta, the active match orange. The zoomed shot is the point of the Custom
Highlight API: the highlights are still exactly on the glyphs after a re-render at a different
scale, with no coordinate code involved. The document shown is the freedesktop.org
shared-mime-info specification, a public document shipped by the `shared-mime-info` package.

### Testing

**A debug APK is attached to [this release](https://github.com/JuliusBairaktaris/PdfViewer/releases/tag/search-preview)** so the PR can be tried
without building it: [PdfViewer-search-debug.apk](https://github.com/JuliusBairaktaris/PdfViewer/releases/download/search-preview/PdfViewer-search-debug.apk). It is debug
signed and installs alongside a normal install as `PDF Viewer d`.

`PdfViewerSearchMatcherTest` (21 tests) pins the matcher's behaviour, which is the one part whose
semantics come from the platform: exact `(start, length)` pairs for case, NFC and NFD diacritics,
ligatures, full width, ß, CJK, non-overlapping matches, cross-line phrase matches, and termination
on a wholly collation-ignorable pattern.

`PdfViewerSearchTest` (5 tests) covers it end to end on a real WebView: counting and wrap-around
navigation across pages, that the Custom Highlight API actually paints ranges covering the matched
word on pdf.js's transformed `color: transparent` text layer, that highlights survive zoom and
rotation, the zero-match state, and that closing clears everything.

`search.test.js` (10 vitest cases) pins the end-of-line rule, including the invariant the whole
offset scheme rests on: every item contributes exactly `str.length + (hasEOL ? 1 : 0)` characters.

Everything below was measured on an API 36 x86_64 emulator running WebView 133 (the minimum this
app accepts), not reasoned about:

- The Custom Highlight API does paint on pdf.js's transformed, `color: transparent` spans, the
  painted range covers the matched word, and highlights re-establish across zoom and rotation.
  This was the riskiest assumption in the design and is now a test, not a hope.
- ICU `PRIMARY` folds case, NFC and NFD diacritics, ligatures, full width, ß and CJK, with exact
  `(start, length)` pairs pinned per case.
- Empty `hasEOL` items are not a corner case: on the 19-page spec used for the screenshots, 187 of
  2128 text items are `{"str": "", "hasEOL": true}`, and each one produces a `textDivs` entry with
  no text node. The null guard in `applyHighlights` runs constantly.
- ICU pays a large one-time cost on its first substantial scan (~12s on this emulator, ~1s per
  244k characters after). Matching is per page, off the UI thread, cancellable and incremental, so
  it shows up as the progress bar taking longer rather than as a stall.
- The full instrumentation suite is green: 114 tests, of which 26 are new and the rest are the
  project's pre-existing ones, unchanged.

### Notes for review

- **Text is retained for the document's lifetime**, capped at 8M characters (~16 MB UTF-16), after
  which the counter reads `n/m+`. That is the cost of not re-sweeping the pdf.js worker on every
  keystroke. If any retention is unwanted in a hardened viewer, dropping the corpus in `onStop` is
  a small change that trades it for a re-scan.
- **Line breaks.** A line-final hyphen becomes two soft hyphens (completely ignorable in root
  collation) so `hyphen-\nation` is found by `hyphenation`; every other line end becomes one space,
  because `U+000A` is *not* collation-equal to a space. This is why a phrase spanning a line break
  is findable.
- **RTL is a known ceiling, not a regression.** pdf.js bidi-reorders `item.str` into *visual* order
  before we ever see it, so a logically-typed Arabic or Hebrew query cannot match a pure-RTL line.
  pdf.js's own find controller has exactly the same limitation. Fixing it means a per-item offset
  transform; happy to add it if wanted.
- Deliberately not included: match-case, whole-word and regex toggles (each is roughly one line in
  the engine plus UI surface and a persisted preference), matches spanning a page boundary, and
  reading-order reconstruction for multi-column layouts. None of these are things pdf.js or
  Chromium do either.
- The 45 vendored `.textLayer .highlight` rules in `text_layer.css` are now permanently unused.
  Left in place to avoid diverging from upstream pdf.js; happy to delete them for a smaller diff.
- **Known ceiling, deliberately not capped.** A hostile PDF with a single page carrying ~1M
  characters can produce the per-page cap of 100k matches for a one-letter query, which is roughly
  a 1.4 MB argument to `evaluateJavascript` and 100k live `Range`s. That is renderer-side jank for
  a fraction of a second, not an ANR, a crash or a wrong result — the Compose UI stays responsive
  because the payload is built off the main thread and shipped asynchronously. Capping it would
  make the counter disagree with what is painted, which seemed the worse trade. Real dense pages
  are three orders of magnitude below this.
- Text extraction visits every page, so pdf.js retains a `PDFPageProxy` per page for the life of
  the document. `pdfDoc.cleanup()` after the sweep would release it; not done here because it
  interacts with the render cache and wanted a maintainer's opinion first.

*Disclaimer: this PR was implemented with the assistance of Claude.*

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -19,4 +19,10 @@ jobs:
           cache: gradle
 
       - name: Build with Gradle
-        run: ./gradlew build --no-daemon
+        run: ./gradlew build --parallel --build-cache --no-daemon
+
+      - name: Upload Debug APK
+        uses: actions/upload-artifact@v4
+        with:
+          name: debug-apk
+          path: app/build/outputs/apk/debug/app-debug.apk
--- a/app/src/androidTest/kotlin/app/grapheneos/pdfviewer/test/PdfViewerSearchMatcherTest.kt
+++ b/app/src/androidTest/kotlin/app/grapheneos/pdfviewer/test/PdfViewerSearchMatcherTest.kt
@@ -1,0 +1,204 @@
+package app.grapheneos.pdfviewer.test
+
+import androidx.test.ext.junit.runners.AndroidJUnit4
+import app.grapheneos.pdfviewer.search.DocumentSearch
+import org.junit.Assert.assertEquals
+import org.junit.Assert.assertFalse
+import org.junit.Assert.assertNotEquals
+import org.junit.Assert.assertTrue
+import org.junit.Test
+import org.junit.runner.RunWith
+
+/**
+ * Pins the behaviour of the ICU collation matcher, which is the one part of search whose
+ * semantics come from the platform rather than from this codebase.
+ */
+@RunWith(AndroidJUnit4::class)
+class PdfViewerSearchMatcherTest {
+
+    private val search = DocumentSearch()
+
+    /** Matches as (start, length) pairs. */
+    private fun find(text: String, pattern: String): List<Pair<Int, Int>> =
+        search.findMatches(text, pattern).toList().chunked(2).map { it[0] to it[1] }
+
+    private fun starts(text: String, pattern: String) = find(text, pattern).map { it.first }
+
+    @Test
+    fun findsPlainSubstring() {
+        assertEquals(listOf(4 to 4), find("the MIME database", "MIME"))
+    }
+
+    @Test
+    fun ignoresCase() {
+        assertEquals(listOf(4 to 4), find("the mime database", "MIME"))
+        assertEquals(listOf(4 to 4), find("the MIME database", "mime"))
+    }
+
+    @Test
+    fun findsEveryOccurrence() {
+        assertEquals(listOf(0, 8, 16), starts("cat bat cat mat cat", "cat"))
+    }
+
+    @Test
+    fun matchesAreNotOverlapping() {
+        assertEquals(listOf(0, 2), starts("aaaa", "aa"))
+    }
+
+    @Test
+    fun reportsNothingWhenAbsent() {
+        assertEquals(emptyList<Pair<Int, Int>>(), find("hello world", "zzz"))
+        assertEquals(emptyList<Pair<Int, Int>>(), find("ab", "abcdef"))
+    }
+
+    @Test
+    fun foldsDiacritics() {
+        assertEquals(listOf(3 to 6), find("Le Résumé final", "resume"))
+        assertEquals(listOf(4 to 6), find("the resume of", "Résumé"))
+    }
+
+    @Test
+    fun foldsDecomposedDiacritics() {
+        // "Résumé" written NFD: e + U+0301 for each accent.
+        val nfd = "Le Résumé final"
+        val hits = find(nfd, "resume")
+        assertEquals(1, hits.size)
+        assertEquals(3, hits[0].first)
+        // Length is in original-string units, so it covers the combining marks too.
+        assertTrue("length ${hits[0].second} should cover the combining marks", hits[0].second >= 6)
+    }
+
+    @Test
+    fun foldsLigatures() {
+        assertEquals(listOf(4 to 1), find("the ﬁle is here", "fi"))
+    }
+
+    @Test
+    fun foldsFullWidth() {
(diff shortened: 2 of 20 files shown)


mio-19:
Github: This branch has conflicts that must be resolved

JuliusBairaktaris:
@thestinger Ready for review.

PR #598, opened 2026-03-03
Add page persistence

### LINKED ISSUE #280: PDF Page Persistence
It would be really nice if it was possible have the PDF viewer automatically open the last opened pdf and open to the last viewed page. Possibly as a toggleble feature.

ggtlvkma356:
Added the following functionalities:

 - Re-open last opened document when cold starting PdfViewer
 - Automatically jump to last opened page of the corresponding document
   - Last page is remembered based on file hash, thus will survive move/copy
   - Up to 50 pairs of last opened page of their corresponding document are saved
 - A simple setting dialog to toggle this behavior on or off

There is an edge case when a document is opened from `ACTION_VIEW` (e.g. from Files) and then the app is killed, then it cannot re-open last document. This is due to a security limitation of Android system.

closes #280


app/src/main/java/app/grapheneos/pdfviewer/PdfViewer.java
```diff
@@ -130,6 +135,8 @@ public class PdfViewer extends AppCompatActivity implements LoaderManager.Loader
     private String mEncryptedDocumentPassword;
     private List<CharSequence> mDocumentProperties;
     private InputStream mInputStream;
+    private PdfPreferencesRepository repository;
+    private Job hashCalculationJob;
```
inthewaves: I think `repository` and `hashCalculationJob` along with the logic for calculating the hash could be put into the PdfViewModel (https://github.com/GrapheneOS/PdfViewer/blob/main/app/src/main/java/app/grapheneos/pdfviewer/viewModel/PdfViewModel.kt) so that we don't need to deal with using Kotlin coroutine code in Java. Then non-suspend methods can be exposed in the ViewModel for PdfViewer.java to call. This will also make it easier for when we eventually decide to rewrite PdfViewer.java into Kotlin.

If you need a Context in the ViewModel for the `repository`, for now I think we can change its subclass to https://developer.android.com/reference/androidx/lifecycle/AndroidViewModel. In the future, we can introduce dependency injection instead to make unit testing easier.

app/src/main/java/app/grapheneos/pdfviewer/KtUtils.kt
```diff
@@ -18,3 +31,115 @@ fun applySystemBarMargins(view: View, applyBottom: Boolean = false) {
...
+): String? = withContext(Dispatchers.IO) {
+    try {
+        contentResolver.openInputStream(uri)?.use { inputStream ->
+            val digest = MessageDigest.getInstance("SHA-256")
+            val buffer = ByteArray(8192)
+            var bytesRead: Int
+
+            while (inputStream.read(buffer).also { bytesRead = it } != -1) {
+                digest.update(buffer, 0, bytesRead)
+            }
```
inthewaves: Looking at it now, I'm wondering how this performs on large PDFs like >= 300 MB. Maybe we could do something like set a limit on bytes hashed (10 KB), and then append file size as part of the key.

app/src/main/java/app/grapheneos/pdfviewer/preferences/PdfPreferencesRepository.kt
```diff
@@ -0,0 +1,119 @@
...
+            preferences[PreferencesKeys.LAST_OPENED_PAGE] = page
+        }
+    }
+
+    suspend fun clearLastOpened() {
+        context.dataStore.edit { preferences ->
+            preferences.remove(PreferencesKeys.LAST_OPENED_URI)
+            preferences.remove(PreferencesKeys.LAST_OPENED_PAGE)
+        }
+    }
```
inthewaves, line 76: We should also release the persistable URI permission here if there was one

app/src/main/java/app/grapheneos/pdfviewer/preferences/PdfPreferencesRepository.kt
```diff
@@ -0,0 +1,119 @@
...
+    suspend fun saveLastOpened(uri: String?, page: Int) {
+        context.dataStore.edit { preferences ->
+            if (uri != null) {
+                preferences[PreferencesKeys.LAST_OPENED_URI] = uri
+            } else {
+                preferences.remove(PreferencesKeys.LAST_OPENED_URI)
+            }
+            preferences[PreferencesKeys.LAST_OPENED_PAGE] = page
+        }
+    }
```
inthewaves, line 69: We should also release the previous persistable URI permission here if there was one

app/src/main/java/app/grapheneos/pdfviewer/KtUtils.kt
```diff
@@ -18,3 +31,115 @@ fun applySystemBarMargins(view: View, applyBottom: Boolean = false) {
...
+    lifecycleOwner: LifecycleOwner,
+    uri: Uri,
+    contentResolver: ContentResolver,
+    onHashCalculated: (String?) -> Unit
+): Job {
+    return lifecycleOwner.lifecycleScope.launch {
+        val hash = calculateFileHash(uri, contentResolver)
+        onHashCalculated(hash)
+    }
+}
```
inthewaves: I think when we move this logic to be in the Kotlin ViewModel, it'd be more idomatic to use `lifecycleOwner.lifecycleScope.async` (or `viewModelScope`) and have this return `Deferred<String>`

inthewaves (changes requested):

app/src/main/java/app/grapheneos/pdfviewer/KtUtils.kt
inthewaves: Actually it probably depends on if we still need the callback, we'll see

app/src/main/java/app/grapheneos/pdfviewer/PdfViewer.java
ggtlvkma356: I have moved all these logic into the view model. Everything shall be more Kotlin-idiomatic and much cleaner now.

app/src/main/java/app/grapheneos/pdfviewer/KtUtils.kt
ggtlvkma356: I changed it to hash only the first 16 KB, shall be very fast.

app/src/main/java/app/grapheneos/pdfviewer/preferences/PdfPreferencesRepository.kt
ggtlvkma356: Fixed in view model.

app/src/main/java/app/grapheneos/pdfviewer/preferences/PdfPreferencesRepository.kt
ggtlvkma356: Also fixed in view model.

app/src/main/java/app/grapheneos/pdfviewer/KtUtils.kt
ggtlvkma356: I have changed the logic to use `Deferred`, no more callbacks needed.

app/src/main/java/app/grapheneos/pdfviewer/viewModel/PdfViewModel.kt
```diff
@@ -90,4 +115,140 @@ class PdfViewModel : ViewModel() {
...
+            // Release old permission if it's different from new URI
+            if (oldUri != null && oldUri != uriString) {
+                releaseUriPermissionIfHeld(oldUri.toUri())
+            }
+
+            if (includeHashMapping && currentFileHash != null) {
+                repository.updatePagePosition(currentFileHash!!, page)
+            }
+        }
+    }
```
inthewaves, line 178: I think this repeated block could be refactored into a suspend function

```kotlin
            val oldState = repository.pdfStateFlow.first()
            val oldUri = oldState.lastOpenedUri

            repository.saveLastOpened(uriString, page)

            // Release old permission if it's different from new URI
            if (oldUri != null && oldUri != uriString) {
                releaseUriPermissionIfHeld(oldUri.toUri())
            }

            if (includeHashMapping && currentFileHash != null) {
                repository.updatePagePosition(currentFileHash!!, page)
            }
```

app/src/main/java/app/grapheneos/pdfviewer/PdfViewer.java
```diff
@@ -153,13 +153,15 @@ public class PdfViewer extends AppCompatActivity implements LoaderManager.Loader
...
                     viewModel.clearOutline();
-                    if (mUri != null) {
-                        takePersistableUriPermission();
-                        savePdfState(false, false);
-                        startHashCalculation();
+                    if (mUri != null && PreferenceHelper.INSTANCE.isResumeLastDocumentEnabled(this)) {
+                        if (takePersistableUriPermission()) {
+                            viewModel.savePdfState(mUri.toString(), mPage, false);
+                        }
+                        viewModel.calculateHash(mUri);
```
inthewaves: We could probably move this repeated block into the ViewModel, maybe as a function:

```
                if (takePersistableUriPermission()) {
                    viewModel.savePdfState(mUri.toString(), mPage, false);
                }
                viewModel.calculateHash(mUri);
```

`takePersistableUriPermission` can also be moved into the ViewModel. Since the ViewModel now handles clearing the URI permissions, I think it's better to have all the persistable URI permission handling in one place.

app/src/main/java/app/grapheneos/pdfviewer/PdfViewer.java
```diff
@@ -542,38 +525,35 @@ public void onZoomEnd() {
...
         } else {
-            if (mUri != null) {
-                takePersistableUriPermission();
-                savePdfState(false, false);
-                startHashCalculation();
+            if (mUri != null && PreferenceHelper.INSTANCE.isResumeLastDocumentEnabled(this)) {
+                if (takePersistableUriPermission()) {
+                    viewModel.savePdfState(mUri.toString(), mPage, false);
+                }
+                viewModel.calculateHash(mUri);
```
inthewaves: See other comment about refactoring this repeated block

app/src/main/java/app/grapheneos/pdfviewer/PdfViewer.java
```diff
@@ -542,38 +525,35 @@ public void onZoomEnd() {
...
+                        try {
+                            getContentResolver().openInputStream(uri).close();
+                            mUri = uri;
+                            mPage = state.getLastOpenedPage();
+                            mNumPages = 0;
+                            viewModel.calculateHash(uri);
+                        } catch (SecurityException | IOException e) {
+                            viewModel.clearLastOpened();
+                            Toast.makeText(this,
+                                    "Previous document is no longer accessible. Please reopen it.",
```
inthewaves: We should put this in strings.xml

app/src/main/java/app/grapheneos/pdfviewer/viewModel/PdfViewModel.kt
```diff
@@ -90,4 +115,140 @@ class PdfViewModel : ViewModel() {
             outline.postValue(if (hasOutline) OutlineStatus.Available else OutlineStatus.NoOutline)
         }
     }
+
+    /**
+     * Load initial PDF state if preference is enabled
+     */
+    fun maybeLoadPdfBlocking(): PdfPreferencesRepository.PdfState? {
```
inthewaves: Name could be `maybeLoadPdfStateBlocking` to be clearer

app/src/main/java/app/grapheneos/pdfviewer/PdfViewer.java
```diff
@@ -46,14 +46,19 @@
 import java.util.Arrays;
 import java.util.HashMap;
 import java.util.List;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
```
inthewaves: These seem unused

inthewaves (changes requested):

app/src/main/java/app/grapheneos/pdfviewer/PdfViewer.java
ggtlvkma356: Fixed in latest.

app/src/main/java/app/grapheneos/pdfviewer/viewModel/PdfViewModel.kt
ggtlvkma356: I have extracted both to a common suspend function.

app/src/main/java/app/grapheneos/pdfviewer/PdfViewer.java
ggtlvkma356: I moved the whole section and also `hasUriPermission` to view model.

app/src/main/java/app/grapheneos/pdfviewer/PdfViewer.java
ggtlvkma356: Same as above.

app/src/main/java/app/grapheneos/pdfviewer/viewModel/PdfViewModel.kt
ggtlvkma356: Fixed.

app/src/main/java/app/grapheneos/pdfviewer/PdfViewer.java
ggtlvkma356: Fixed.

inthewaves (approved):

app/src/main/java/app/grapheneos/pdfviewer/KtUtils.kt
```diff
@@ -17,4 +17,4 @@ fun applySystemBarMargins(view: View, applyBottom: Boolean = false) {
         v.layoutParams = mlp
         windowInsets
     }
-}
+}
```
quh4gko8: minor code nit, remove this empty change

ggtlvkma356: Removed.

app/src/main/java/app/grapheneos/pdfviewer/fragment/SettingsDialog.kt
```diff
@@ -0,0 +1,43 @@
...
+import android.os.Bundle
+import androidx.core.content.edit
+import androidx.fragment.app.DialogFragment
+import app.grapheneos.pdfviewer.R
+import app.grapheneos.pdfviewer.databinding.DialogSettingsBinding
+import com.google.android.material.dialog.MaterialAlertDialogBuilder
+
+class SettingsDialog : DialogFragment() {
+
+    private var _binding: DialogSettingsBinding? = null
```
muhomorr: This property is redundant, it's used only inside onCreate()

app/src/main/java/app/grapheneos/pdfviewer/fragment/SettingsDialog.kt
```diff
@@ -0,0 +1,43 @@
...
+    private val binding get() = _binding!!
+
+    override fun onCreateDialog(savedInstanceState: Bundle?): Dialog {
+        _binding = DialogSettingsBinding.inflate(layoutInflater)
+
+        val prefs = requireContext().getSharedPreferences("app_prefs", Context.MODE_PRIVATE)
```
muhomorr: PreferenceHelper.PREF_NAME

app/src/main/java/app/grapheneos/pdfviewer/hashing/FileHash.kt
```diff
@@ -0,0 +1,49 @@
...
+import android.net.Uri
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.withContext
+import java.io.IOException
+import java.security.MessageDigest
+
+object FileHash {
+
+    private const val HASH_BUFFER_SIZE = 8192
+    private const val MAX_HASH_BYTES = 16 * 1024  // 16KB
```
muhomorr: Why size limit is this low? Are first 16 KiB of a PDF file guaranteed to be distinct for all PDF files?

app/src/main/java/app/grapheneos/pdfviewer/hashing/FileHash.kt
```diff
@@ -0,0 +1,49 @@
...
+                while (totalBytesRead < MAX_HASH_BYTES &&
+                    inputStream.read(buffer).also { bytesRead = it } != -1) {
+
+                    val bytesToHash = minOf(bytesRead, MAX_HASH_BYTES - totalBytesRead)
+
+                    digest.update(buffer, 0, bytesToHash)
+                    totalBytesRead += bytesToHash
+                }
+
+                digest.digest().joinToString("") { "%02x".format(it) }
```
muhomorr: Need to use kotlin.text.HexFormat for such conversions, the current approach is very inefficient

app/src/main/java/app/grapheneos/pdfviewer/viewModel/PdfViewModel.kt
```diff
@@ -90,4 +116,157 @@ class PdfViewModel : ViewModel() {
...
+     *
+     * @param uri Document URI
+     */
+    fun calculateHash(uri: Uri) {
+        hashCalculationDeferred?.cancel()
+        currentFileHash = null
+
+        hashCalculationDeferred = viewModelScope.async {
+            val hash = FileHash.calculateFileHash(uri, contentResolver)
+            currentFileHash = hash
```
muhomorr: There is a race condition: currentFileHash write is not guaranteed to be cancelled when hashCalculationDeferred is cancelled

app/src/main/java/app/grapheneos/pdfviewer/preferences/PdfPreferencesRepository.kt
```diff
@@ -0,0 +1,119 @@
...
+
+            preferences[PreferencesKeys.FILE_PAGE_POSITIONS] = serializeFilePagePositions(currentMap)
+        }
+    }
+
+    suspend fun getPageForFile(fileHash: String): Int? {
+        return pdfStateFlow.first().filePagePositions[fileHash]
+    }
+
+    suspend fun clearAll() {
```
muhomorr: Why is this function unused?

app/src/main/java/app/grapheneos/pdfviewer/preferences/PdfPreferencesRepository.kt
```diff
@@ -0,0 +1,119 @@
...
+    private fun parseFilePagePositions(json: String?): Map<String, Int> {
+        if (json.isNullOrEmpty()) return emptyMap()
+        return try {
+            val jsonObject = JSONObject(json)
+            buildMap {
+                jsonObject.keys().forEach { key ->
+                    put(key, jsonObject.getInt(key))
+                }
+            }
+        } catch (e: Exception) {
```
muhomorr: This conversion should never fail, it's a bug if it does fail. We shouldn't hide potential bugs by swallowing exceptions

app/src/main/java/app/grapheneos/pdfviewer/viewModel/PdfViewModel.kt
```diff
@@ -90,4 +116,157 @@ class PdfViewModel : ViewModel() {
...
+            )
+            return true
+        } catch (_: SecurityException) {
+            return false
+        }
+    }
+
+    fun hasUriPermission(uri: Uri?): Boolean {
+        return contentResolver.persistedUriPermissions
+            .stream()
```
muhomorr: Why are you using Java stream in Kotlin code?

app/src/main/java/app/grapheneos/pdfviewer/PdfViewer.java
```diff
@@ -498,9 +519,35 @@ public void onZoomEnd() {
...
+                viewModel.prepareNewPdf(mUri, mPage);
+            } else {
+                PdfPreferencesRepository.PdfState state =
+                        viewModel.maybeLoadPdfStateBlocking();
+
+                if (state != null && state.getLastOpenedUri() != null) {
+                    Uri uri = Uri.parse(state.getLastOpenedUri());
+                    if (viewModel.hasUriPermission(uri)) {
+                        try {
+                            getContentResolver().openInputStream(uri).close();
```
muhomorr: Opening a Uri is an expensive operation, especially for some third-party content providers. Content resolver result should be reused

app/src/main/java/app/grapheneos/pdfviewer/viewModel/PdfViewModel.kt
```diff
@@ -90,4 +116,157 @@ class PdfViewModel : ViewModel() {
...
+        if (takePersistableUriPermission(uri)) {
+            savePdfState(uri.toString(), page, false)
+        }
+        calculateHash(uri)
+    }
+
+    private fun releaseUriPermissionIfHeld(uri: Uri) {
+        try {
+            // Find if we have persisted permission for this URI
+            val persistedPermissions = contentResolver.persistedUriPermissions
```
muhomorr: It's unnecessary and slow to obtain all persisted Uri permissions.

This entire code block can be replaced with 

```
try {
    contentResolver.releasePersistableUriPermission(uri, Intent.FLAG_GRANT_READ_URI_PERMISSION or Intent.FLAG_GRANT_WRITE_URI_PERMISSION)   
} catch (e: SecurityException) {
    Log.w(TAG, "permission release failed", e)
}
```

muhomorr (changes requested):

app/src/main/java/app/grapheneos/pdfviewer/fragment/SettingsDialog.kt
ggtlvkma356: I removed the property.

app/src/main/java/app/grapheneos/pdfviewer/fragment/SettingsDialog.kt
ggtlvkma356: Fixed.

app/src/main/java/app/grapheneos/pdfviewer/hashing/FileHash.kt
ggtlvkma356: ~I expand it to 10 MB. I looked it up and on modern mobile devices this shall take at most 1 second, which sounds fine to me.~

app/src/main/java/app/grapheneos/pdfviewer/hashing/FileHash.kt
ggtlvkma356: Fixed.

app/src/main/java/app/grapheneos/pdfviewer/viewModel/PdfViewModel.kt
ggtlvkma356: I moved the assignment to when the `Deferred` is actually used.

app/src/main/java/app/grapheneos/pdfviewer/preferences/PdfPreferencesRepository.kt
ggtlvkma356: Removed.

app/src/main/java/app/grapheneos/pdfviewer/preferences/PdfPreferencesRepository.kt
ggtlvkma356: I removed the try..catch block.

app/src/main/java/app/grapheneos/pdfviewer/viewModel/PdfViewModel.kt
ggtlvkma356: It was moved from the Java main class. I changed it to use Kotlin `any`.

app/src/main/java/app/grapheneos/pdfviewer/PdfViewer.java
ggtlvkma356: I changed it to just rely on the permission check, so no need to open the actual uri.

app/src/main/java/app/grapheneos/pdfviewer/viewModel/PdfViewModel.kt
ggtlvkma356: Fixed.

app/src/main/java/app/grapheneos/pdfviewer/hashing/FileHash.kt
ggtlvkma356: The hash is removed. It now uses fingerprint from `pdf.js` directly, which [reads ID from a pdf file first, and if that fails, computes MD5 hash for the first 1KB](https://github.com/mozilla/pdf.js/blob/9d81fafa8c4783f3026057cbbf096722a42c841a/src/core/document.js#L1610).

app/src/main/java/app/grapheneos/pdfviewer/preferences/PdfPreferencesRepository.kt
```diff
@@ -0,0 +1,110 @@
...
+import androidx.datastore.preferences.preferencesDataStore
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.catch
+import kotlinx.coroutines.flow.first
+import kotlinx.coroutines.flow.map
+import org.json.JSONObject
+
+private val Context.dataStore: DataStore<Preferences> by preferencesDataStore(
+    name = "pdf_preferences"
+)
```
inthewaves, line 22: I think we should also handle data format corruption here (corruptionHandler) to explicitly replace it with empty preferences. We can keep the `.catch` in the flow for other errors like actual IO issues during reading

inthewaves (changes requested):

app/src/main/java/app/grapheneos/pdfviewer/preferences/PdfPreferencesRepository.kt
ggtlvkma356: Yes, it is a good idea. I added a `ReplaceFileCorruptionHandler` to clear the preferences when there is a corruption. The exception handler for reading can be kept the same for `IOException`.

gradle/verification-metadata.xml
```diff
@@ -3,6 +3,11 @@
    <configuration>
       <verify-metadata>true</verify-metadata>
       <verify-signatures>false</verify-signatures>
+      <trusted-artifacts>
+         <trust group="gradle" name="gradle"/>
+         <trust file=".*-javadoc[.]jar" regex="true"/>
+         <trust file=".*-sources[.]jar" regex="true"/>
```
inthewaves: `<trust file=".*-src[.]zip" regex="true"/>` should cover the Gradle zip in Android Studio, and we should document this

```suggestion
         <!-- https://issuetracker.google.com/issues/340494392#comment2 -->
         <trust file=".*-src[.]zip" regex="true"/>
         <trust file=".*-javadoc[.]jar" regex="true"/>
         <trust file=".*-sources[.]jar" regex="true"/>
```

mio-19:
github: This branch has conflicts that must be resolved

PR #603, opened 2026-03-09
Allow jumping to previous/next page with horizontal swipe gesture

### LINKED ISSUE #41: gesture-based navigation

ggtlvkma356:
This is an improved version of #383, which lacks edge detection, in which case page will change regardless if we are in the middle of a page or at its edge.

The swipe is horizontal only.

closes #41.

### DIFF
--- a/app/src/main/java/app/grapheneos/pdfviewer/GestureHelper.java
+++ b/app/src/main/java/app/grapheneos/pdfviewer/GestureHelper.java
@@ -7,6 +7,11 @@ import android.view.MotionEvent;
 import android.view.ScaleGestureDetector;
 import android.view.View;
 
+import androidx.annotation.NonNull;
+import androidx.annotation.Nullable;
+
+import java.util.concurrent.atomic.AtomicBoolean;
+
 /*
     The GestureHelper present a simple gesture api for the PdfViewer
 */
@@ -14,6 +19,7 @@ import android.view.View;
 class GestureHelper {
     public interface GestureListener {
         boolean onTapUp();
+        boolean onFling(@Nullable MotionEvent e1, @NonNull MotionEvent e2, float velocityX, float velocityY);
         void onZoom(float scaleFactor, float focusX, float focusY);
         void onZoomEnd();
     }
@@ -21,18 +27,12 @@ class GestureHelper {
     @SuppressLint("ClickableViewAccessibility")
     static void attach(Context context, View gestureView, GestureListener listener) {
 
-        final GestureDetector detector = new GestureDetector(context,
-                new GestureDetector.SimpleOnGestureListener() {
-                    @Override
-                    public boolean onSingleTapUp(MotionEvent motionEvent) {
-                        return listener.onTapUp();
-                    }
-                });
+        AtomicBoolean wasScaling = new AtomicBoolean(false);
 
         final ScaleGestureDetector scaleDetector = new ScaleGestureDetector(context,
                 new ScaleGestureDetector.SimpleOnScaleGestureListener() {
                     @Override
-                    public boolean onScale(ScaleGestureDetector detector) {
+                    public boolean onScale(@NonNull ScaleGestureDetector detector) {
                         listener.onZoom(detector.getScaleFactor(), detector.getFocusX(),
                                 detector.getFocusY());
 
@@ -40,14 +40,44 @@ class GestureHelper {
                     }
 
                     @Override
-                    public void onScaleEnd(ScaleGestureDetector detector) {
+                    public void onScaleEnd(@NonNull ScaleGestureDetector detector) {
                         listener.onZoomEnd();
                     }
                 });
 
+        final GestureDetector detector = new GestureDetector(context,
+                new GestureDetector.SimpleOnGestureListener() {
+                    @Override
+                    public boolean onSingleTapUp(@NonNull MotionEvent motionEvent) {
+                        return listener.onTapUp();
+                    }
+
+                    @Override
+                    public boolean onFling(@Nullable MotionEvent e1, @NonNull MotionEvent e2,
+                                           float velocityX, float velocityY) {
+                        if (wasScaling.get()) {
+                            return false;
+                        }
+
+                        return listener.onFling(e1, e2, velocityX, velocityY);
+                    }
+                });
+
         gestureView.setOnTouchListener((view, motionEvent) -> {
+            int action = motionEvent.getActionMasked();
+
+            if (action == MotionEvent.ACTION_DOWN) {
+                wasScaling.set(false);
+            }
+
             detector.onTouchEvent(motionEvent);
             scaleDetector.onTouchEvent(motionEvent);
+
+            // Check after scaleDetector processes the event
+            if (scaleDetector.isInProgress()) {
+                wasScaling.set(true);
+            }
+
             return false;
         });
     }
--- a/app/src/main/java/app/grapheneos/pdfviewer/PdfViewer.java
+++ b/app/src/main/java/app/grapheneos/pdfviewer/PdfViewer.java
@@ -13,7 +13,9 @@ import android.view.Gravity;
 import android.view.Menu;
 import android.view.MenuInflater;
 import android.view.MenuItem;
+import android.view.MotionEvent;
 import android.view.View;
+import android.view.ViewConfiguration;
 import android.webkit.CookieManager;
 import android.webkit.JavascriptInterface;
 import android.webkit.RenderProcessGoneDetail;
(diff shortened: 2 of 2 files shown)


app/src/main/java/app/grapheneos/pdfviewer/PdfViewer.java
```diff
@@ -450,6 +454,43 @@ public boolean onTapUp() {
...
+                        if (e1 == null) return false;
+
+                        float deltaX = e2.getX() - e1.getX();
+                        float deltaY = e2.getY() - e1.getY();
+                        float absDeltaX = Math.abs(deltaX);
+                        float absDeltaY = Math.abs(deltaY);
+
+                        float screenDensity = getResources().getDisplayMetrics().density;
+                        int swipeThreshold = (int) (SWIPE_THRESHOLD * screenDensity);
+                        int swipeVelocityThreshold = (int) (SWIPE_VELOCITY_THRESHOLD * screenDensity);
```
inthewaves: We should precompute `swipeThreshold` and `swipeVelocityThreshold` instead of calculating it every fling

inthewaves (changes requested):

app/src/main/java/app/grapheneos/pdfviewer/PdfViewer.java
ggtlvkma356: I changed it to precomputed, also now the thresholds use system-defined values.

app/src/main/java/app/grapheneos/pdfviewer/GestureHelper.java
```diff
@@ -45,6 +41,29 @@ public void onScaleEnd(ScaleGestureDetector detector) {
...
+                    public boolean onFling(@Nullable MotionEvent e1, @NonNull MotionEvent e2,
+                                           float velocityX, float velocityY) {
+                        if (scaleDetector.isInProgress()) {
+                            return false;
+                        }
+
+                        // Ignore multi-touch
+                        if (e1 != null && (e1.getPointerCount() > 1 || e2.getPointerCount() > 1)) {
+                            return false;
+                        }
```
inthewaves: Android delivers touch events one at a time. When the user lifts their fingers from a pinch, this always happens as two separate events, one per finger (physically impossible to lift both fingers in the exact same instant)

The first finger to leave the screen produces a ACTION_POINTER_UP event. At this point, the `ScaleGestureDetector` sees that only one finger remains, so it ends the scale gesture and sets `isInProgress()` to false. however, `GestureDetector` does nothing special on ACTION_POINTER_UP;  it never checks for flings on this event type. It only checks for flings when the _last_ finger leaves the screen.

The second (last) finger to leave produces a different ACTION_UP event. Now `GestureDetector` wakes up: it checks the remaining finger's velocity, and if it's fast enough, fires `onFling`. But the scale already ended on the previous event. By the time `onFling` runs and checks `isInProgress()`, the answer is always false, as the scale ended one event ago.

This means the `isInProgress()` guard only works if both fingers are still touching the screen when `onFling` fires. But that never happens because `onFling` requires the last finger to lift (ACTION_UP), and the scale requires at least two fingers to stay active. The scale always ends first.

The same two-event split explains why the pointer count check also doesn't work as intended. The `e1` event (the original ACTION_DOWN) was recorded when the first finger touched so that's one pointer. The `e2` event (the ACTION_UP) is recorded when the last finger lifts so that's also one pointer, because the other finger already left on the earlier ACTION_POINTER_UP. `GestureDetector` never updates `e1` when a second finger arrives, and `e2` only ever sees the one finger that remains. So both `e1.getPointerCount()` and `e2.getPointerCount()` are `1`, and the multi-touch guard sees a gesture that looks entirely single-finger.

Can add these logs to GestureHelper to confirm this:

```java
@Override
public boolean onFling(@Nullable MotionEvent e1, @NonNull MotionEvent e2,
                        float velocityX, float velocityY) {
    Log.d(TAG, "onFling: isInProgress=" + scaleDetector.isInProgress()
            + " e1.pointers=" + (e1 != null ? e1.getPointerCount() : "null")
            + " e2.pointers=" + e2.getPointerCount()
            + " vX=" + velocityX + " vY=" + velocityY);
```

```java
gestureView.setOnTouchListener((view, motionEvent) -> {
    final int action = motionEvent.getActionMasked();
    final boolean scaleInProgress = scaleDetector.isInProgress();

    if (action == MotionEvent.ACTION_DOWN) {
        Log.d(TAG, "ACTION_DOWN: pointers=" + motionEvent.getPointerCount());
    } else if (action == MotionEvent.ACTION_POINTER_DOWN) {
        Log.d(TAG, "ACTION_POINTER_DOWN: pointers=" + motionEvent.getPointerCount() + " isInProgress=" + scaleInProgress);
    } else if (action == MotionEvent.ACTION_POINTER_UP) {
        Log.d(TAG, "ACTION_POINTER_UP: pointers=" + motionEvent.getPointerCount() + " isInProgress=" + scaleInProgress);
    } else if (action == MotionEvent.ACTION_UP) {
        Log.d(TAG, "ACTION_UP: pointers=" + motionEvent.getPointerCount() + " isInProgress=" + scaleInProgress);
    }
```

10:51:17 is when I put both fingers down, 10:51:20 is when I released both fingers during a zoom gesture at the side. In terms of `onFling`, `e1` happens at 10:51:17.887 (only 1 pointer), and `e2` happens at 10:51:20.992.

```
2026-04-07 10:51:17.887  7592-7592  GestureHelper           app.grapheneos.pdfviewer.debug       D  ACTION_DOWN: pointers=1                            // This is e1 from onFling
2026-04-07 10:51:17.893  7592-7592  GestureHelper           app.grapheneos.pdfviewer.debug       D  ACTION_POINTER_DOWN: pointers=2 isInProgress=false
2026-04-07 10:51:20.991  7592-7592  GestureHelper           app.grapheneos.pdfviewer.debug       D  ACTION_POINTER_UP: pointers=2 isInProgress=true    // Scale ends after this event
2026-04-07 10:51:20.992  7592-7592  GestureHelper           app.grapheneos.pdfviewer.debug       D  ACTION_UP: pointers=1 isInProgress=false           // This is e2, i.e., what triggers the fling
2026-04-07 10:51:20.992  7592-7592  GestureHelper           app.grapheneos.pdfviewer.debug       D  onFling: isInProgress=false e1.pointers=1 e2.pointers=1 vX=145.1329 vY=67.527855
```

We could add a `mWasScaling`  to `GestureHelper` inside of ` gestureView.setOnTouchListener` and set that to true when `scaleDetector.isInProgress()`. This `mWasScaling` can replace both the scaleDetector and pointer count checks. Then only reset `mWasScaling` to false when user taps again, i.e., ACTION_DOWN is received

app/src/main/java/app/grapheneos/pdfviewer/PdfViewer.java
```diff
@@ -125,6 +128,8 @@ public class PdfViewer extends AppCompatActivity implements LoaderManager.Loader
     private float mZoomRatio = 1f;
     private float mZoomFocusX = 0f;
     private float mZoomFocusY = 0f;
+    private int swipeThreshold;
+    private int swipeVelocityThreshold;
```
inthewaves: nit: these should be prefixed with m

app/src/main/java/app/grapheneos/pdfviewer/PdfViewer.java
```diff
@@ -520,6 +560,12 @@ public void onZoomEnd() {
         }
     }
 
+    private void initializeGestures() {
+        ViewConfiguration vc = ViewConfiguration.get(this);
+        swipeThreshold = vc.getScaledTouchSlop() * 4;
```
inthewaves: I think it's a bit too sensitive, maybe 6 would work better

inthewaves (changes requested):

app/src/main/java/app/grapheneos/pdfviewer/PdfViewer.java
inthewaves: Actually I saw the other PR, this is probably okay

ggtlvkma356: Yeah this is based on the assumption that it will be on top of #616.

Edit: no longer applicable.

app/src/main/java/app/grapheneos/pdfviewer/PdfViewer.java
ggtlvkma356: Adjusted to 6 now.

app/src/main/java/app/grapheneos/pdfviewer/GestureHelper.java
ggtlvkma356: Thanks for the detailed explanation and demonstrations. Indeed this is something I missed and it is good to learn how touch events work exactly.

I have updated the implementation to use a `wasScaling` to track state in `setOnTouchListener`, shall work correctly now.

inthewaves (approved):

thestinger:
Needs to be rebased.

PR #616, opened 2026-03-27
Multiple improvements and refactors

ggtlvkma356:
## List of changes

1. Move UI operations to the main thread.

2. Make shared mutable state thread-safe.

3. Move file saving off UI thread.

4. Remove Loader. Document properties are loaded using ViewModel.

5. Remove `onSaveInstanceState`. Persistent states are managed with ViewModel.

6. Simplify Menu handling.

7. Display page number with Snackbar instead of Toast.

8. Remove Hungarian notation.


app/src/main/java/app/grapheneos/pdfviewer/PdfViewer.java
```diff
@@ -841,38 +771,28 @@ private void saveDocument() {
...
-            if (property.toString().startsWith("Title:-")) {
-                title = property.toString().replace("Title:-", "");
-            }
-        }
-        return fileName.length() > 2 ? fileName : title;
+        Map<DocumentProperty, String> properties = viewModel.getDocumentProperties().getValue();
+        if (properties == null || properties.isEmpty()) return "";
+        String fileName = properties.getOrDefault(DocumentProperty.FileName, "");
+        String title = properties.getOrDefault(DocumentProperty.Title, "");
+        return !(fileName != null && fileName.isEmpty()) ? fileName : title;
```
inthewaves: Expanding it right now would be `fileName == null || !fileName.isEmpty()` so it will accept null if that's intentional. (although null never happens  because of `properties.getOrDefault`

We can consider also just using !TextUtils.isEmpty. `TextUtils.isEmpty` is a framework method that returns true iff the string is null or 0-length

app/src/main/java/app/grapheneos/pdfviewer/viewModel/PdfViewModel.kt
```diff
@@ -90,4 +154,49 @@ class PdfViewModel : ViewModel() {
...
+        viewModelScope.launch(Dispatchers.IO) {
+            val loader = DocumentPropertiesRetriever(getApplication(), properties, numPages, uri)
+            val result = loader.retrieve()
+            withContext(Dispatchers.Main) {
+                if (documentPropertiesLoading) {
+                    _documentProperties.value = result
+                }
+            }
+        }
+        documentPropertiesLoading = true
```
inthewaves: I think `documentPropertiesLoading = true` should be before `viewModelScope.launch(Dispatchers.IO)`. Technically doesn't matter since this seems to be on main thread anyway, so `withContext(Dispatchers.Main)` would always catch this

app/src/main/java/app/grapheneos/pdfviewer/PdfViewer.java
```diff
@@ -114,28 +97,18 @@ public class PdfViewer extends AppCompatActivity implements LoaderManager.Loader
...
-    private List<CharSequence> mDocumentProperties;
-    private InputStream mInputStream;
+
+    private final Object streamLock = new Object();
+
+    private volatile float zoomFocusX = 0f;
+    private volatile float zoomFocusY = 0f;
+    private boolean documentLoaded;
+    private volatile InputStream inputStream;
+    private volatile boolean documentPropertiesLoaded;
```
inthewaves: I don't think volatile works here for this pattern. For this compound operation of read-and-then-set, it only helps with read visibility

```java
@JavascriptInterface
public void setDocumentProperties(final String properties) {
    if (documentPropertiesLoaded) {
        throw new SecurityException("setDocumentProperties already called");
    }
    documentPropertiesLoaded = true;
    ...
}
```

https://docs.oracle.com/javase/tutorial/essential/concurrency/atomic.html

> Atomic actions cannot be interleaved, so they can be used without fear of thread interference. However, this does not eliminate all need to synchronize atomic actions, because memory consistency errors are still possible.

AtomicBoolean#compareAndSet would be more correct here.

Just as a note from my own looking: even if this bridge method itself is synchronous with respect to JS code, `documentPropertiesLoaded` is still technically read/written by Android main thread in `loadPdf()`, so doesn't seem correct to drop this pattern entirely

app/src/main/java/app/grapheneos/pdfviewer/PdfViewer.java
```diff
@@ -821,6 +750,7 @@ public boolean onOptionsItemSelected(MenuItem item) {
             return true;
         } else if (itemId == R.id.action_save_as) {
             saveDocument();
+            return true;
```
inthewaves, line 877: Good catch

inthewaves (changes requested):
As a general note: Technically, JavaScript is single-threaded and the JS Bridge code blocks when calling bridge methods, so concurrency between bridge calls is less of an issue (i.e., JS will never concurrently call bridge methods). Of course, the JS bridge background thread in Java still can have concurrency issue with pure Android threads

From Chromium source on [`android_webview/docs/java-bridge.md`](https://source.chromium.org/chromium/chromium/src/+/main:an): *"Calls to interface methods are synchronous—JavaScript VM stops and waits for a result to be returned from the invoked method. In Chromium, this means that the IPC message sent from a renderer to the browser must be synchronous."*

app/src/main/java/app/grapheneos/pdfviewer/viewModel/PdfViewModel.kt
```diff
@@ -90,4 +158,75 @@ class PdfViewModel : ViewModel() {
...
+            val loader = DocumentPropertiesRetriever(getApplication(), properties, numPages, uri)
+            val result = loader.retrieve()
+            val name = resolveDocumentName(result)
+            withContext(Dispatchers.Main) {
+                if (documentPropertiesLoading) {
+                    _documentProperties.value = result
+                    _documentName.value = name
+                }
+            }
+        }
```
inthewaves: There's technically a race here if document property loading takes a long time. If you load properties for a doc and then switch to another doc and the previous job is still running, it could replace it with previous doc properties

When a new document is loaded, I think this job should probably be cancelled, or at the very least the job should check that the input `uri` parameter matches the `uri` before replacing `_documentProperties.value` and `_documentName.value`

Note that `DocumentPropertiesRetriever` and its usage of `PDFJsPropertiesToDocumentPropertyConverter` are not suspending functions and they don't call cancellable functions (they seem to be pure Kotlin functions), so cancelling the job would only stop the job after it has finished converting when it enters `withContext(Dispatchers.Main)`

app/src/main/java/app/grapheneos/pdfviewer/PdfViewer.java
```diff
@@ -982,39 +898,35 @@ private void saveDocument() {
...
+        Uri oldUri = viewModel.getUri();
+        if (oldUri != null) {
+            try {
+                getContentResolver().releasePersistableUriPermission(oldUri, Intent.FLAG_GRANT_READ_URI_PERMISSION);
+            } catch (SecurityException ignored) {}
         }
+        try {
+            getContentResolver().takePersistableUriPermission(newUri, Intent.FLAG_GRANT_READ_URI_PERMISSION);
+        } catch (SecurityException ignored) {}
+        viewModel.setUri(newUri);
```
inthewaves: I think this is from the page persistence branch, we should remove it

app/src/main/java/app/grapheneos/pdfviewer/PdfViewer.java
```diff
@@ -201,27 +169,29 @@ public class PdfViewer extends AppCompatActivity implements LoaderManager.Loader
...
-            return mZoomRatio;
+            return viewModel.getZoomRatio();
         }
 
         @JavascriptInterface
         public void setZoomRatio(final float ratio) {
-            mZoomRatio = Math.max(Math.min(ratio, MAX_ZOOM_RATIO), MIN_ZOOM_RATIO);
+            runOnUiThread(() ->
+                    viewModel.setZoomRatio(Math.max(Math.min(ratio, MAX_ZOOM_RATIO), MIN_ZOOM_RATIO))
+            );
```
inthewaves: Note that this pattern technically makes setting the zoom ratio asynchronous which changes the semantics. JS Bridge code should block JS exectuion when calling bridge methods, so offloading to Android main thread could result in later bridge `getZoomRatio` calls having a stale value

This is probably unlikely in practice depending on how far away subsequent `getZoomRatio` calls are from `setZoomRatio` calls. But I think  we can just remove this `runOnUiThread` call and rely on `zoomRatio` being volatile

inthewaves (changes requested):

app/src/main/java/app/grapheneos/pdfviewer/viewModel/PdfViewModel.kt
ggtlvkma356: I think the final results are still correct in this case, because if the first document property is still loading when you open a second document, it will start a new coroutine and eventually will overwrite the results from the first one (unless the second coroutine somehow finishes before the first one, which is very unlikely). But it is good to cancel the job in any case so I will add the cancellation.

app/src/main/java/app/grapheneos/pdfviewer/PdfViewer.java
ggtlvkma356: Yes the `runOnUiThread` is removed now.

app/src/main/java/app/grapheneos/pdfviewer/viewModel/PdfViewModel.kt
ggtlvkma356: This is no longer needed after the job cancellation changes below.

app/src/main/java/app/grapheneos/pdfviewer/viewModel/PdfViewModel.kt
```diff
@@ -1,15 +1,74 @@
...
+            field = value
+            savedStateHandle[STATE_DOCUMENT_ORIENTATION_DEGREES] = value
+        }
 
-class PdfViewModel : ViewModel() {
+    @Volatile
+    var numPages: Int = 0
+
+    @Volatile
+    var encryptedDocumentPassword: String = ""
```
inthewaves, line 69: `encryptedDocumentPassword` is not stored in some saved state so it won't be restored after process death, but I think that's actually fine to not keep a password around for that long. Still saved after activity recreation like rotation, etc.

inthewaves (approved):
Looks fine, instrumented tests passed locally

This should be squashed when it's merged

app/src/main/java/app/grapheneos/pdfviewer/viewModel/PdfViewModel.kt
ggtlvkma356: Yes, it is changed intentionally so that passwords are not saved after process is dead.

thestinger:
@ggtlvkma356 Can you rebase this yourself to reduce the number of commits?

ggtlvkma356:
I have squashed the commits into one.

inthewaves (approved):

PR #627, opened 2026-04-24
Add instrumented tests and js tests

ggtlvkma356:
Adds instrumented tests that covers most of the user interactions.

The JS tests only cover outline, because `pdf.js` also has their own tests, which covers most cases already.

The changes to production code is minimal:
 - Changed visibility of two variables in main activity from private to package private
 - Moved one JS function to a separate file


app/src/androidTest/kotlin/app/grapheneos/pdfviewer/test/PdfViewerNavigationTest.kt
```diff
@@ -0,0 +1,197 @@
...
+            PdfViewerTestUtils.waitForCanvasRendered(scenario)
+
+            scenario.onActivity {
+                assertEquals("Page should reset to 1", 1, it.currentPage)
+                assertEquals("New document should have 1 page", 1, it.totalPages)
+            }
+
+            PdfViewerTestUtils.assertTextLayerContent(scenario, "Test Text")
+        }
+    }
```
inthewaves, line 277: This test format is great: loading PDF A, doing some stuff, then load PDF B and verify all state is reset

Currently, this only tests that the page is reset, but there's other state could assert for as well (e.g., zoom state, rotation)

Also we could add some other test cases too, like 

- loading test-multipage.pdf which has an outline, checking that outline is visible in menu, then load some other PDF that doesn't have an outline, and ensure outline option is no longer shown in menu (and vice versa)
- opening non-encrypted PDF first, then open encrypted PDF and ensure it still asks for password

app/src/androidTest/java/app/grapheneos/pdfviewer/test/TestPdfProvider.java
```diff
@@ -0,0 +1,96 @@
...
+                os.close();
+                is.close();
+            } catch (IOException e) {
+                throw new FileNotFoundException(
+                        "Failed to copy asset '" + filename + "': " + e.getMessage());
+            }
+        }
+
+        return ParcelFileDescriptor.open(file, ParcelFileDescriptor.MODE_READ_ONLY);
+    }
```
inthewaves, line 62: It might be worth adding a test for content provider open failure in middle of reading file (e.g. opening document from cloud provider or USB gets unplugged). Could do this like a query parameter on the Uri that would make it throw an exception partway

inthewaves (changes requested):
Coverage looks good, just some gaps we could add

inthewaves (changes requested):
Also, I'm sure this seems a bit trivial (I think so as well), but it could be good for reproducibility to include the .tex files used to generate the PDFs somewhere (or put them all in one markdown in separate code blocks)

app/src/androidTest/kotlin/app/grapheneos/pdfviewer/test/PdfViewerNavigationTest.kt
ggtlvkma356: I have added these tests. However, for rotation, current it is not reset after loading a second pdf, so the assertion is commented out. I think this is a bug. The rest works correctly.

app/src/androidTest/java/app/grapheneos/pdfviewer/test/TestPdfProvider.java
ggtlvkma356: I added the test, but for the current implementation it doesn't do much, because the UI didn't show error if I'm not mistaken. Adding some feedback will be a separate PR.

ggtlvkma356:
> Also, I'm sure this seems a bit trivial (I think so as well), but it could be good for reproducibility to include the .tex files used to generate the PDFs somewhere (or put them all in one markdown in separate code blocks)

I have added a Markdown file in the `assets` folder to include all three latex files.

app/src/androidTest/java/app/grapheneos/pdfviewer/test/TestPdfProvider.java
inthewaves: We can open an issue for this error reporting then

app/src/androidTest/kotlin/app/grapheneos/pdfviewer/test/PdfViewerNavigationTest.kt
inthewaves: We can open an issue for this then

ggtlvkma356:
I opened #635, #636 to track the two issues.

inthewaves (approved):

PR #631, opened 2026-04-27
Support PDF 2.0 dates, fix time zone and validation handling

### LINKED ISSUE #629: accept PDF 2.0 (and 1.7) dates
https://developer.adobe.com/document-services/docs/assets/5b15559b96303194340b99820d3a70fa/PDF_ISO_32000-2.pdf has date format `D:YYYYMMDDHHmmSSOHH'mm` and includes this example and note:

>For example, December 23, 1998, at 7:52 PM, U.S. Pacific Standard Time, is represented by the string:
>
>     D:199812231952-08'00
>
> NOTE PDF versions up to and including 1.7 defined a date string to include a terminating apostrophe.
> PDF processors are recommended to accept date strings that still follow that convention.

Contrast with PDF 1.7 format at https://opensource.adobe.com/dc-acrobat-sdk-docs/pdfstandards/pdfreference1.7old.pdf being `D:YYYYMMDDHHmmSSOHH'mm'`:

> For example, December 23, 1998, at 7:52 PM, U.S. Pacific Standard Time, is represented by the string
> 
>     D:199812231952−08'00'

thomasbuilds:
closes #629

### DIFF
--- a/app/src/main/java/app/grapheneos/pdfviewer/Utils.java
+++ b/app/src/main/java/app/grapheneos/pdfviewer/Utils.java
@@ -1,14 +1,45 @@
 package app.grapheneos.pdfviewer;
 
-import android.text.TextUtils;
-
 import java.math.RoundingMode;
 import java.text.DateFormat;
 import java.text.DecimalFormat;
 import java.text.ParseException;
 import java.util.Calendar;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 public class Utils {
+
+    // PDF date string format. Based on the format described in section 7.9.4 of
+    // the PDF 32000-2:2020 specification:
+    //
+    //     D:YYYYMMDDHHmmSSOHH'mm
+    //
+    // The PDF 1.7 reference defined the same format with a terminating
+    // apostrophe, and PDF processors are recommended to accept date strings
+    // that follow that older convention. The apostrophe between HH and mm is
+    // also tolerated as missing for additional leniency, matching pdf.js.
+    private static final Pattern PDF_DATE_PATTERN = Pattern.compile(
+            "^D:" +
+                    "(\\d{4})" +     // Year (required)
+                    "(\\d{2})?" +    // Month (optional)
+                    "(\\d{2})?" +    // Day (optional)
+                    "(\\d{2})?" +    // Hours (optional)
+                    "(\\d{2})?" +    // Minutes (optional)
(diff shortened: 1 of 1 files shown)


thomasbuilds:
I noticed a couple of preexisting issues in `parseDate`. ~~They may be worth handling in a separate PR~~, since this PR is mainly about accepting the PDF 2.0 date syntax, but they’re relevant to the rewritten parser.

**1. UTC offset handling is semantically inverted.**

The current arithmetic, preserved from `main`, does:

```java
case "-":
    hours -= offsetHours;
    minutes -= offsetMinutes;
    break;
case "+":
    hours += offsetHours;
    minutes += offsetMinutes;
    break;
```

Per ISO 32000-2 §7.9.4, an offset of `-08'00` means the local time is 8 hours earlier than UTC, so normalizing to UTC should add 8 hours, not subtract. pdf.js handles it the other way:

https://github.com/mozilla/pdf.js/blob/v5.6.205/src/display/display_utils.js#L551-L561

However, just flipping the signs would not fully fix it, because `Calendar.getInstance()` uses the device timezone. If we normalize fields to UTC and then set them on a default-timezone `Calendar`, the resulting instant is still wrong. A complete fix would probably need to build the calendar in UTC, then format the resulting `Date` in the default timezone, for example

```java
final Calendar calendar = Calendar.getInstance(TimeZone.getTimeZone("UTC"));
// set fields and apply offset to normalize to UTC
return DateFormat.getDateTimeInstance(...).format(calendar.getTime());
```

**2. `Calendar` leniency accepts impossible dates.**

`D:20230231` is currently accepted and silently rolls over to March 3. If we want strict validation, we’d need something like

```java
calendar.setLenient(false);
calendar.getTime();
```

ggtlvkma356:
For the time calculation, what about setting timezone on calendar?

```
TimeZone tz = TimeZone.getTimeZone(String.format("GMT%s%02d:%02d", sign, hours, minutes));
calendar.setTimeZone(tz);
```

and then

```
DateFormat.getDateTimeInstance().format(calendar.getTime())
```

thomasbuilds:
refactored to modern java.time

ggtlvkma356:
The `java.time` approach is much cleaner, lgtm.

@inthewaves It needs an approval to merge, or @thestinger can merge it directly?

inthewaves (approved):

PR #640, opened 2026-05-05
Adapt security policies for workers

### LINKED ISSUE #637: Missing CSP, Permission policy, etc. on workers
CSP and Permission Policy are not applied to `worker.js`. Need to investigate and enforce similar rules.

ggtlvkma356:
Fixes #637.

Tested with #627.

### DIFF
--- a/app/src/main/java/app/grapheneos/pdfviewer/PdfViewer.java
+++ b/app/src/main/java/app/grapheneos/pdfviewer/PdfViewer.java
@@ -76,8 +76,16 @@ public class PdfViewer extends AppCompatActivity implements LoaderManager.Loader
         "img-src blob: 'self'; " +
         "script-src 'self'; " +
         "style-src 'self'; " +
+        "worker-src 'self'; " +
         "frame-ancestors 'none'; " +
         "base-uri 'none'";
+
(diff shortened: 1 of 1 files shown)


app/src/main/java/app/grapheneos/pdfviewer/PdfViewer.java
```diff
@@ -76,9 +76,17 @@ public class PdfViewer extends AppCompatActivity implements LoaderManager.Loader
         "img-src blob: 'self'; " +
         "script-src 'self'; " +
         "style-src 'self'; " +
+        "worker-src 'self'; " +
```
inthewaves, line 79: I guess this was technically handled by worker-src falling back to script-src before: https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy/worker-src. It's good that we're now explicit about it though

inthewaves (approved):

app/src/main/java/app/grapheneos/pdfviewer/PdfViewer.java
ggtlvkma356: Yes it is technically not needed but I think making it explicit is good.

PR #648, opened 2026-05-08
Add workflows for tests

ggtlvkma356:


thestinger:
It may be better to rename the build workflow to build-and-test with separate jobs to run as we've done here:

https://github.com/GrapheneOS/hardened_malloc/blob/main/.github/workflows/build-and-test.yml

You can see it creates multiple separate checks which run separately from the hardened_malloc CI output.

inthewaves:
I believe third-party action reactivecircus/android-emulator-runner also needs to be allowed in repo settings when this is merged. This does follow practices for third-party actions at https://docs.github.com/en/actions/reference/security/secure-use#writing-workflows and https://docs.github.com/en/actions/reference/security/secure-use#using-third-party-actions

.github/workflows/build-and-test.yml
```diff
@@ -0,0 +1,36 @@
...
+
+permissions:
+  contents: read
+
+jobs:
+  instrumented-tests:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
```
inthewaves, line 40: Since we're not making authenticated Git commands, we could harden the checkout action further by specifying 

```
with:
  persist-credentials: false
```

https://github.com/actions/checkout#checkout-v4

> The auth token is persisted in the local git config. This enables your scripts to run authenticated git commands. The token is removed during post-job cleanup. Set `persist-credentials: false` to opt-out.

In checkout v6:

> Improved credential security: `persist-credentials` now stores credentials in a separate file under `$RUNNER_TEMP` instead of directly in `.git/config`

Though I think this would apply to existing checkout actions in other workflows too, maybe in a separate commit or a different PR. The read-only permission for the token above also helps with token issues, but doesn't hurt to also use `persist-credentials: false`

.github/workflows/build-and-test.yml
```diff
@@ -0,0 +1,36 @@
+name: Android Instrumented tests
+
+on: [pull_request, push]
+
+permissions:
+  contents: read
```
inthewaves, line 6: This could probably be good for the other workflows too, in another commit or another PR.

I can confirm that https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#permissions does say this applies to all jobs in this workflow (`instrumented-tests.yml`) when used as a top-level key, and it does restrict it to just `read`:

> For each of the available permissions, shown in the table below, you can assign one of the access levels: `read` (if applicable), `write`, or `none`. `write` includes `read`. If you specify the access for any of these permissions, all of those that are not specified are set to `none`.

inthewaves (changes requested):

.github/workflows/build-and-test.yml
ggtlvkma356: I applied it to the gradle validation workflow too.

.github/workflows/build-and-test.yml
ggtlvkma356: Yes, it is a good addition. I applied it to all workflows now.

ggtlvkma356:
> It may be better to rename the build workflow to build-and-test with separate jobs to run as we've done here:
> 
> https://github.com/GrapheneOS/hardened_malloc/blob/main/.github/workflows/build-and-test.yml

I have updated the workflow. Now it runs three separate jobs in parallel inside one workflow, similar to the one linked.

ggtlvkma356:
> I believe third-party action reactivecircus/android-emulator-runner also needs to be allowed in repo settings when this is merged. This does follow practices for third-party actions at https://docs.github.com/en/actions/reference/security/secure-use#writing-workflows and https://docs.github.com/en/actions/reference/security/secure-use#using-third-party-actions

In fact, because all actions are pinned to commits instead of versions now, the repo settings need to be updated to allow specific commits for them instead. Github doesn't seem to match smartly. See the error messages in https://github.com/GrapheneOS/PdfViewer/actions/runs/25680517388 and https://github.com/GrapheneOS/PdfViewer/actions/runs/25680514505.

inthewaves:
Was thinking about the minutes this would consume  (e.g. https://github.com/ggtlvkma356/PdfViewer/actions/runs/25680513571 takes 10 minutes), but this should be fine, since GitHub runners are free for public repos using standard GitHub-hosted runners: https://docs.github.com/en/billing/concepts/product-billing/github-actions

ggtlvkma356:
Yes, the standard runners are free for public repos. There is a 6 hours limit on how long they can run, but we are way below that so this is not a problem.

inthewaves (approved):

PR #664, opened 2026-06-03
Update workflows and tests for CI

ggtlvkma356:
- Update workflows for better emulator environment
- Increase timeouts for polling operations in tests
- Use JUnit per-test retry rule instead of full workflow retries

This removes the need for retry the whole flows for multiple times and is generally more stable.


inthewaves (approved):
LGTM

PR #668, opened 2026-06-05
Upgrade dependencies

ggtlvkma356:
Supersede #653, #658, #661, #665

* Bump com.google.android.material:material from 1.13.0 to 1.14.0
* Bump com.google.devtools.ksp:symbol-processing-gradle-plugin from 2.3.7 to 2.3.9
* Bump org.jetbrains.kotlin:kotlin-gradle-plugin from 2.3.21 to 2.4.0
* Bump androidx.core:core-ktx from 1.18.0 to 1.19.0


thestinger:
I think KSP doesn't officially support Kotlin 2.4.0 yet based on their new version numbering so it may make sense to use 2.3.20 until it's officially supported. However, I don't know if it's actually used in practice since we don't use KSP ourselves and only added it to override the version alongside the Kotlin version since it gets pulled in by default.

ggtlvkma356:
@thestinger KSP versions are now untied with Kotlin versions, see https://github.com/google/ksp/pull/2658. The official get start page shows the combination of KSP 2.3.9 with Kotlin 2.4.0: https://kotlinlang.org/docs/ksp-quickstart.html#add-a-ksp-based-processor-to-your-project.

thestinger:
Ah, that's good.

PR #692, opened 2026-07-13
Zoom menu item and other minor fixes

### LINKED ISSUE #569: Desktop Mode zoom
From what I understand, there is currently no way to zoom in and out on a document when using PdfViewer in desktop mode. It would be great to add the ability to do so, since the pdfs tend to open very zoomed out. Separate zoom buttons in the menu and/or ctrl + scroll wheel would be great to have.

ggtlvkma356:
- Add a menu item for zooming, and a dialog for entering numbers when clicked
- Fix wrong text color on WebView crash screen
- Update icons to use Material Symbols

Closes #569


app/src/main/java/app/grapheneos/pdfviewer/PdfViewerScreen.kt
```diff
@@ -946,39 +1007,166 @@ private fun WebViewAlertScreen(
...
-                style = MaterialTheme.typography.titleLarge,
-                modifier = Modifier.padding(top = 8.dp)
+        }
+
+        TextButton(
+            onClick = onCustomZoom,
+            enabled = enabled,
+            modifier = Modifier.testTag(TestTags.ZOOM_PERCENTAGE)
+        ) {
+            Text("${(zoomRatio * 100).roundToInt()}%")
```
inthewaves: Seems better to use java.text.NumberFormat which has `getPercentInstance` for percentage formatting

app/src/main/java/app/grapheneos/pdfviewer/PdfViewerScreen.kt
```diff
@@ -946,39 +1007,166 @@ private fun WebViewAlertScreen(
...
-            Text(
-                text = title,
-                style = MaterialTheme.typography.titleLarge,
-                modifier = Modifier.padding(top = 8.dp)
+        }
+
+        TextButton(
+            onClick = onCustomZoom,
+            enabled = enabled,
+            modifier = Modifier.testTag(TestTags.ZOOM_PERCENTAGE)
```
inthewaves, line 1086: Could add `.semantics { liveRegion = LiveRegionMode.Polite }` for accessibility

app/src/main/java/app/grapheneos/pdfviewer/PdfViewerScreen.kt
```diff
@@ -946,39 +1007,166 @@ private fun WebViewAlertScreen(
...
+                                text = filtered,
+                                selection = TextRange(
+                                    minOf(newValue.selection.end, filtered.length)
+                                )
+                            )
+                        }
+                    },
+                    singleLine = true,
+                    isError = textFieldValue.text.isNotEmpty() && !isValid,
+                    suffix = { Text("%") },
```
inthewaves: There are some locales where the `%` is placed at the front but is still LTR, e.g. Turkish. Although we don't have general localization support, it might be better to be more future proof

app/src/main/java/app/grapheneos/pdfviewer/PdfViewerScreen.kt
```diff
@@ -349,8 +366,10 @@ fun PdfViewerScreen(
             }
 
             override fun onZoom(scaleFactor: Float, focusX: Float, focusY: Float) {
-                viewModel.zoomRatio = (viewModel.zoomRatio * scaleFactor)
-                    .coerceIn(MIN_ZOOM_RATIO, MAX_ZOOM_RATIO)
+                viewModel.setZoomRatio(
+                    (viewModel.zoomRatio.value * scaleFactor)
+                        .coerceIn(MIN_ZOOM_RATIO, MAX_ZOOM_RATIO)
+                )
                 viewModel.zoomFocusX = focusX
```
inthewaves, line 382: Seems to be an issue with cached zooming, e.g.

1. Set zoom to 200% in the dialog
2. Press preset to increase to 300%, then go back to 200%
3. Scroll away from the page origin, then begin pinching outward to zoom in
4. Observe the page jump or pan toward the wrong location as the pinch begins instead of retaining the content beneath the fingers

https://github.com/user-attachments/assets/c1ec3f35-06f9-47fc-82bb-032dba148c7e

app/src/main/java/app/grapheneos/pdfviewer/PdfViewerScreen.kt
```diff
@@ -220,13 +235,15 @@ fun PdfViewerScreen(
     val documentProperties by viewModel.documentProperties.collectAsStateWithLifecycle()
     val outlineStatus by viewModel.outline.collectAsStateWithLifecycle()
     val showPasswordDialog by viewModel.showPasswordDialog.collectAsStateWithLifecycle()
+    val zoomRatio by viewModel.zoomRatio.collectAsStateWithLifecycle()
```
inthewaves: This seems to be causing the `PdfTopAppBar` to recompose itself on every single zoom ratio change; simply zooming in and out caused it to recompose 100s of times

This can be observed in the Layout Inspector in Android Studio using Running Devices button on the right menu > Press + to add device > Top right to toggle Layout Inspector, then do pinch zooming

`ScaleGestureDetector.onScale()` runs repeatedly throughout the gesture. Each distinct ratio now:

1. updates the `StateFlow`;
2. schedules recomposition from the screen-level state read;
3. passes a changed `Float` into `PdfTopAppBar`, preventing that subtree from being skipped;
4. independently issues `evaluateJavascript("onRenderPage(2)")`.

This work occurs even when the overflow menu is closed and the zoom percentage is not visible. It theoretically adds Compose work on the same main thread hot path responsible for interactive pinch rendering

We should observe the zoom flow at the narrowest subtree that displays it, ideally inside dropdown content that is only composed while expanded

inthewaves (changes requested):

app/src/main/java/app/grapheneos/pdfviewer/PdfViewerScreen.kt
ggtlvkma356: The formatting is using `NumberFormat.getPercentInstance` now.

app/src/main/java/app/grapheneos/pdfviewer/PdfViewerScreen.kt
ggtlvkma356: Added.

app/src/main/java/app/grapheneos/pdfviewer/PdfViewerScreen.kt
ggtlvkma356: I used a boolean switch for this one and get the direction from `NumberFormat`. The most correct solution seems to be using `VisualTransformation` but for this simple input that is probably overkilling.

app/src/main/java/app/grapheneos/pdfviewer/PdfViewerScreen.kt
ggtlvkma356: This is fixed now.

app/src/main/java/app/grapheneos/pdfviewer/PdfViewerScreen.kt
ggtlvkma356: You are right, I didn't realize the compose is happening so many times. The flow is moved to be inside `ZoomRow` now.

inthewaves (approved):

PR #702, opened 2026-07-17
Dependency update

ggtlvkma356:
- Bump com.android.application from 9.2.1 to 9.3.0 (#698)
- Bump org.jetbrains.kotlin.plugin.compose from 2.4.0 to 2.4.10 (#697)
- Bump org.jetbrains.kotlin:kotlin-gradle-plugin from 2.4.0 to 2.4.10 (#696)
- Bump org.jetbrains.kotlin.plugin.serialization from 2.4.0 to 2.4.10 (#695)


thestinger:
@ggtlvkma356 This should be done by removing `gradle/verification-metadata.xml` and using the script from https://github.com/GrapheneOS/app-docs/tree/main/scripts now. We want to get the legacy entries which aren't required anymore stripped out over time and use this script to include the additional files used by Android Studio and other software instead of only what Gradle thinks is needed.

ggtlvkma356:
@thestinger I have updated `verification-metadata.xml` with the `resolveDependencyVerificationArtifacts` command using that script.

PR #721, opened 2026-08-14
Let zoom-out menu reach the true minimum zoom

xhon-pelushi:
## Summary
- Addresses https://github.com/GrapheneOS/PdfViewer/issues/711

## Test plan
- See worklog in contributor notes; tested locally where feasible.

### DIFF
--- a/app/src/main/java/app/grapheneos/pdfviewer/PdfViewerScreen.kt
+++ b/app/src/main/java/app/grapheneos/pdfviewer/PdfViewerScreen.kt
@@ -144,7 +144,14 @@ private fun nextZoomPreset(ratio: Float): Float? {
 
 private fun previousZoomPreset(ratio: Float): Float? {
     val currentPercent = (ratio * 100).roundToInt()
-    return ZOOM_PRESETS.lastOrNull { it < currentPercent }?.let { it / 100f }
+    val preset = ZOOM_PRESETS.lastOrNull { it < currentPercent }?.let { it / 100f }
+    if (preset != null) {
+        return preset
+    }
+    // No preset below the current zoom: fall back to the true minimum, since pinch-to-zoom
+    // can reach lower than the smallest preset.
+    val minPercent = (MIN_ZOOM_RATIO * 100).roundToInt()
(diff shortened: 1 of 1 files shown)


xhon-pelushi:
Cross-referencing: #736 (@munzzyy) fixes the same bug by a different route, and the two are worth
comparing before either is merged.

**#736** changes the preset list itself:

```kotlin
private val ZOOM_PRESETS = intArrayOf(
    (MIN_ZOOM_RATIO * 100).roundToInt(), 50, 75, 100, 125, 150, 200, 300, 500, 750, 1000
)
```

**This PR** leaves the presets alone and adds a fallback to the true minimum when no preset is
below the current zoom.

The difference is not cosmetic. `MIN_ZOOM_RATIO` is `0.2f` (`PdfJsChannel.kt:9`) and the first
preset today is `25`. So #736 replaces `25` with `20`, which means **the 25% step stops existing** —
zooming out from 50% would go straight to 20% instead of stopping at 25% first. This PR keeps the
ladder intact and only adds a final step: 50% → 25% → 20%.

In #736's favour, and it is a real point: it ships an instrumented test
(`PdfViewerNavigationTest.zoomOutButton_reachesMinZoom`) and this PR has none.

I have not built either branch on a device, so the above is read from the diffs and the constants
rather than observed — take it as a pointer to check rather than a verdict. If you prefer #736's
shape, I am happy to close this; if you prefer keeping the preset ladder, I can add an equivalent
test here.

munzzyy:
You are right, and I checked rather than taking it on trust. MIN_ZOOM_RATIO is 0.2f, so my change turns the first rung from 25 into 20 and the 25% step stops existing. Trading a rung away to reach the minimum is a bad deal.

Keeping the ladder and adding a final step is the better shape: previousZoomPreset returns the last preset below the current value, or the true minimum when there is none. That gives 50 to 25 to 20. My test starts at 0.25 and asserts the button reaches the minimum and then disables, so it still covers it.

I will rework #736 to that. If a maintainer would rather take this PR, close mine; the ladder is the part worth keeping.

PR #736, opened 2026-09-09
Make the zoom out button reach the minimum zoom

### LINKED ISSUE #711: zoom menu item can't be used to reach minimum zoom
It treats 25% as the minimum but the zoom can be made lower via pinch to zoom (currently 20%).

munzzyy:
Fixes #711.

The zoom out button clamps at 25% because ZOOM_PRESETS starts at 25, while pinch to zoom and the custom zoom dialog both go down to MIN_ZOOM_RATIO (20%). This derives the lowest preset from MIN_ZOOM_RATIO so the menu and pinch limits can't drift apart.

Added zoomOutButton_reachesMinZoom next to the existing min/max button tests: set zoom to 25%, click zoom out, assert the ratio lands on MIN_ZOOM_RATIO and the button disables there. It fails without the fix (the button is already disabled at 25%) and passes with it. Ran the full PdfViewerNavigationTest, PdfViewerMenuStateTest and PdfViewerRenderTest classes on an API 36 x86_64 emulator (42 tests passing) plus npm test.


munzzyy:
Reworked as discussed in #721: the presets stay as they were and zoom out falls back to the minimum once no preset is left, so it goes 50% to 25% to 20%. I added a test that 50% still steps to 25%. It fails against my first version, which is the case @xhon-pelushi caught.

